PR #1, opened 2016-03-15
sql: SQL-backed StateMachine running and test

maxtaco:
- virtualize the clock for testing through to SQL
- query builder
- sqlite3 debugging
- abstract date concept for SQLite
- special SQL scanners for reading data out of SQL
- all tests working on SQLite
- debug and test InbandMessagesSince.
- some work on docs


oconnor663:
Do we need any notion of "only one device at a time should handle this?" For example, if KBFS rekey notifications go out to all devices over this system, we might not want all devices to attempt a rekey at the same time. (It sounds like KBFS has a built-in mechanism for making this idempotent, but other callers might not?)

sm_test.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+func consumeMessage(t *testing.T, which string, sm StateMachine, m Message) {
+	err := sm.ConsumeMessage(m)
+	if err != nil {
+		t.Fatalf("In inserting msg %s: %v", which, err)
+	}
+}
+
+func testStateMachineAllDevices(t *testing.T, sm StateMachine, fc clockwork.FakeClock) {
```
maxtaco: add a bunch of comment here

gabriel:
One thing to be careful of is when using time or clocks is that certain devices, like Android phones or tablets, can have weird clock issues and/or devices can straight up lie about their clocks (I've seen iPhones do this too).

One way I've dealt with this in the past is to send down with responses the service clock time and to use that as the clock value or to determine if the clock is wrong on clients and fail gracefully etc.

maxtaco:
Indeed.  The server clock is authoritative and clients should just send
offsets.  Hence the timeoroffset type. And if possible the time will be
generated by MySQL and not the go machines. Hence the time abstraction
class.

PR #3, opened 2016-03-20
build out a protocol file for gregor

maxtaco:
* make tweaks to the original gregor interfaces
* make AVDL interface that we're going to use
* do the first parts of the network infrastructure, but we'll likely punt on that for a bit


keybase/protocol/avdl/common.avdl
```diff
@@ -0,0 +1,71 @@
+@namespace("gregor.1")
+
+protocol common {
+
+	import idl "github.com/keybase/client/go/protocol" as keybase1;
+
+	record TimeOrOffset {
+		keybase1.Time time_;
```
patrickxb: What are these underscore suffixes?

maxtaco: Can't have the member the same name as the method!!!!! You asked for it by getting rid of `GetX` :)

patrickxb: Why do you need methods for returning public struct fields?

maxtaco: To fit the interface.

patrickxb: Could the avdl compiler then translate

    record TimeOrOffset { keybase1.Time time_; }

to 

    struct TimeOrOffset { keybase1.Time time }

    func (t TimeOrOffset) Time() { return t.time }

???

maxtaco: We could, it would be super magical and would be complicated on the other platforms. Is it worth it?

interface.go
```diff
@@ -119,18 +118,39 @@ type ObjFactory interface {
...
-	BrodcastInbandMessage(m InbandMessage) error
-	BrodcastOutOfBandMessage(m OutOfBandMessage) error
 	TriggerNotification(m InbandMessage) error
 }
+
+func UIDFromMessage(m Message) UID {
+	if ibm := m.ToInbandMessage(); ibm != nil {
+		if md := ibm.Metadata(); md != nil {
+			return md.UID()
+		}
```
patrickxb: return nil here to save wasted attempt to convert to out of band message?

keybase/protocol/avdl/common.avdl
patrickxb: We've never had auto-generated methods on protocol records before that satisfy an interface...

Another thought would be to just preserve the case of the record field.

keybase/protocol/go/extras.go
```diff
@@ -0,0 +1,220 @@
...
+		return i.StateSync.Md
+	}
+	return nil
+}
+
+func (i InbandMessage) ToStateSyncMessage() gregor.StateSyncMessage {
+	if i.StateSync == nil {
+		return nil
+	}
+	return i.StateSync
```
patrickxb: seems like this is the only line needed in this function.

keybase/protocol/go/extras.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func (i InbandMessage) ToStateUpdateMessage() gregor.StateUpdateMessage {
+	if i.StateUpdate == nil {
+		return nil
+	}
+	return i.StateUpdate
```
patrickxb: same here, this is the only line needed.

maxtaco: i've often been burned by the nil interface / nil pointer bug.

keybase/protocol/go/extras.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (m Message) ToInbandMessage() gregor.InbandMessage {
+	if m.Ibm == nil {
+		return nil
+	}
+	return m.Ibm
+}
+
+func (m Message) ToOutOfBandMessage() gregor.OutOfBandMessage {
+	if m.Oobm != nil {
```
patrickxb: Did you mean `==` here?

keybase/protocol/go/extras.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (o OutOfBandMessage) Body() gregor.Body     { return o.Body_ }
+func (o OutOfBandMessage) System() gregor.System { return o.System_ }
+func (o OutOfBandMessage) UID() gregor.UID       { return o.Uid }
+
+func (m Message) ToInbandMessage() gregor.InbandMessage {
+	if m.Ibm == nil {
+		return nil
+	}
+	return m.Ibm
```
patrickxb: It seems like you must have a reason for this pattern...

keybase/protocol/go/factory.go
```diff
@@ -0,0 +1,173 @@
...
+
+func castUID(uid gregor.UID) (UID, error) {
+	ret, ok := uid.(UID)
+	if !ok {
+		return UID(""), fmt.Errorf("Bad UID; wrong type")
+	}
+	return ret, nil
+}
+
+func castItem(i gregor.Item) (ItemAndMetadata, error) {
```
patrickxb: rename to `castItemAndMetadata`

keybase/protocol/go/factory.go
```diff
@@ -0,0 +1,173 @@
...
+	ret, ok := i.(ItemAndMetadata)
+	if !ok {
+		return ItemAndMetadata{}, fmt.Errorf("Bad Item; wrong type")
+	}
+	return ret, nil
+}
+
+func castMsgID(msgid gregor.MsgID, e *error) MsgID {
```
patrickxb: Is there a reason some `cast` functions take an error pointer parameter and some return (type, error)?

maxtaco: this is stale code, i will nuke these.  i don't believe it's called anymore

keybase/protocol/go/extras.go
patrickxb: Ok, well see line 171 below...I think that definitely needs fixing.

maxtaco: In particular, what happen if I return a `(*StateUpdate)(nil)` as a `gregor.SateUpdateMessage`?  I think we'll get the wrong answer, and *not* a `nil` interface.

jacobhaven:
I'd be happier with `s/Inband/InBand/g`

maxtaco:
this is a slightly annoying change across a lot of files.  does everyone else agree?

maxtaco:
done

patrickxb:
No preference on the name.

patrickxb:
PR:  LGTM!

PR #5, opened 2016-03-24
if you ask for per-device state, you should also get global state too

maxtaco:


maxtaco:
@oconnor663 can you help me review this

maxtaco:
If you ask for State() for a device, you should get both the per-device messages and the global messages, but previously it was only getting the per-device messages....

patrickxb:
LGTM

PR #6, opened 2016-03-24
docs

maxtaco:


maxtaco:
Don't think there's much need for a review here, ready to just merge the docs and then we can discuss.

PR #11, opened 2016-03-25
Vendor external dependencies, remove libkb dependency

patrickxb:
r? @maxtaco


patrickxb:
Will redo vendor later after maxtaco/CORE-2708 UID/DeviceID changes.

PR #12, opened 2016-03-25
factor out Keybase dependency

maxtaco:
- Just duplicate the protocol, we'll be copying deviceID and UID, but probably worth it to simplify.
- Copy over `Time` implementation
- Some basic beginnings on the server RPC handling.


patrickxb:
LGTM, much simpler.

PR #14, opened 2016-03-26
top-level main server process

maxtaco:
- command-line parsing
- TLS port establishment and listening
- drop into a MainLoop -- TBD behind an interface
- stop on a signal or an explicit stopper
- tests for lots of the flag parsing and basic server establishment
- vendor GoAMZ


server/config.go
```diff
@@ -0,0 +1,279 @@
...
+	return o.TLSOptions.Cert != "" && o.TLSOptions.Key != ""
+}
+
+func (o *Options) BindAddr() string {
+	var ret string
+	if o.BindHost != "" {
+		ret = o.BindHost
+	}
+	ret += ":"
+	ret += fmt.Sprintf("%d", o.Port)
```
jacobhaven: Use [net.JoinHostPort](https://golang.org/pkg/net/#JoinHostPort)

server/config.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+type TLSOptions struct {
+	Key  string
+	Cert string
+}
+
+type Options struct {
+	SessionServer *url.URL
+	Port          int
```
jacobhaven: It's simpler to just have `Options#Port` and `rawOpts#port` as `string`s. Also removes the need for `getEnvInt`.

server/server.go
```diff
@@ -0,0 +1,33 @@
...
+	opts   *Options
+	mls    gregor.MainLoopServer
+	stopCh chan struct{}
+}
+
+func newMainServer(o *Options, m gregor.MainLoopServer) *mainServer {
+	return &mainServer{opts: o, mls: m}
+}
+
+func (m *mainServer) run() error {
```
jacobhaven: `run` would normally be called `ListenAndServe` with
```Go
`MainLoop(n net.Listener) error`
```
being
```Go
Serve(l net.Listener) error
```

server/listen.go
```diff
@@ -0,0 +1,22 @@
...
+	cert, err := tls.X509KeyPair([]byte(m.opts.TLSOptions.Cert), []byte(m.opts.TLSOptions.Key))
+	if err != nil {
+		return nil, err
+	}
+	conf := tls.Config{Certificates: []tls.Certificate{cert}}
+	return tls.Listen("tcp", m.opts.BindAddr(), &conf)
+}
+
+func (m *mainServer) listen() (net.Listener, error) {
+	if m.opts.UseTLS() {
```
jacobhaven: I would prefer instead of `m.opts.UseTLS()`, to have a `m.opts.Config *tls.Config` set by `TLSOptions#Parse`. Then this becomes:
```Go
func (m *mainServer) listen() (net.Listener, error) {
	if m.opts.Config != nil {
		return tls.Listen("tcp", m.opts.BindAddr(), m.opts.Config)
	}
	return net.Listen("tcp", m.opts.BindAddr())
}

File: server/server_test.go [hunk 5]
```diff
@@ -0,0 +1,256 @@
...
+	"strings"
+	"testing"
+)
+
+func writeTmp(d string) (string, error) {
+	f, err := ioutil.TempFile("", "gregor_test")
+	if err != nil {
+		return "", err
+	}
+	src := strings.NewReader(d)
```
jacobhaven: You can simply call [f.WriteString(d)](https://golang.org/pkg/os/#File.WriteString)

server/config.go
```diff
@@ -148,27 +135,21 @@ func (o *Options) UseTLS() bool {
...
-	if raw.port == 0 {
-		return badUsage("No valid listen port specified")
+	if raw.bindAddress == "" {
+		return badUsage("No valid bind-address specified")
 	}
-	o.Port = raw.port
+
 	var err error
+	if _, _, err = net.SplitHostPort(raw.bindAddress); err != nil {
+		return badUsage("bad bind-address: %s", err)
```
jacobhaven: If you want to be strict, you could check that
```Go
if _, err := strconv.Atoi(port); err!=nil {
	return badUsage("bad port (%s): %s", port, err)
}
```

maxtaco: I wrongly supposed that `SplitHostAndPort` would catch it.  Added a test too

server/config.go
```diff
@@ -101,38 +101,34 @@ func readEnvOrFile(name string) (string, error) {
 	return string(res), nil
 }
 
-func (o *TLSOptions) Parse(raw *rawOpts) error {
+func (o *TLSOptions) Parse(raw *rawOpts) (bool, error) {
```
jacobhaven: I think it makes more sense to write this as:
```Go
ParseTLSConfig(raw *rawOpts) (*tls.Config, error)
```

It can parse out an actual `*tls.Config` for you (that can even be directly edited before use), instead of using this structure to generate a new `*tls.Config` each time `listenTLS` is called.

server/config.go
```diff
@@ -141,15 +142,19 @@ func (o *Options) Parse(raw *rawOpts) error {
 		return badUsage("No valid bind-address specified")
 	}
 
-	var err error
-	if _, _, err = net.SplitHostPort(raw.bindAddress); err != nil {
+	if _, port, err := net.SplitHostPort(raw.bindAddress); err != nil {
 		return badUsage("bad bind-address: %s", err)
+	} else if _, err = strconv.Atoi(port); err != nil {
```
jacobhaven: Actually, now that I think of it `strconv.ParseUint(v, 10, 16)` is probably better. It checks that the port is in the range 0-65535.

server/config.go
```diff
@@ -0,0 +1,256 @@
...
+
+type TLSOptions struct {
+	Key  string
+	Cert string
+}
+
+type Options struct {
+	SessionServer *url.URL
+	BindAddress   string
+	MysqlDSN      *url.URL
```
jacobhaven: I could see the argument for replacing this with a [*sql.DB](https://golang.org/pkg/database/sql/#DB), which is basically just a pointer to a database connection pool configuration. No connections are actually opened until it's used.

PR #15, opened 2016-03-28
simple and inefficient in-memory storage

maxtaco:
- useful for when SQLite isn't possible
- uses very naive data structures and lots of linear scans
- serializes all access
- passes all storage tests
- should be good enough for testing

### DIFF
--- a/storage/mem_sm.go
+++ b/storage/mem_sm.go
@@ -1,0 +1,307 @@
+package storage
+
+import (
+	"bytes"
+	"encoding/hex"
+	"github.com/jonboulle/clockwork"
+	gregor "github.com/keybase/gregor"
+	"sync"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


storage/mem_sm.go
```diff
@@ -0,0 +1,307 @@
...
+	gregor "github.com/keybase/gregor"
+	"sync"
+	"time"
+)
+
+// MemEngine is an implementation of a gregor StateMachine that just keeps
+// all incoming messages in a hash table, with one entry per user. It doesn't
+// do anything fancy w/r/t indexing Items, so just iterates over all of them
+// every time a dismissal or a state dump comes in. Used mainly for testing
+// which SQLite isn't available.
```
patrickxb: which -> when

patrickxb:
LGTM

maxtaco:
comment fixed too

PR #16, opened 2016-03-30
RPC machinery

patrickxb:
First cut at implementing the RPC machinery for gregor.

All tests pass, with -race too.  Probably other tests to add, but wanted to get feedback started now.

PTAL @maxtaco


rpc/connection.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if err := conn.startRPCServer(); err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+func (c *connection) Authenticate(ctx context.Context, tok protocol.AuthToken) error {
+	log.Printf("Authenticate: %+v", tok)
```
maxtaco: what's the logging story?

rpc/connection.go
```diff
@@ -0,0 +1,94 @@
...
+	uid        protocol.UID
+	session    protocol.SessionID
+	lastAuthed time.Time
+	parent     *Server
+	authCh     chan error
+}
+
+func newConnection(c net.Conn, parent *Server) (*connection, error) {
+	// TODO: logging and error wrapping mechanisms.
+	xprt := rpc.NewTransport(c, nil, nil)
```
maxtaco: we should handle these todos sooner rather than later.  in particular, we should probably use the convection from our client code that error objects are structs that can be converted to typed Errors unambiguously.

rpc/connection.go
patrickxb: I went with the simplest option for now.  What should we do for gregor logging?

rpc/connection.go
patrickxb: Ok.  Would you like me to do that in this ticket?

rpc/connection.go
maxtaco: https://keybase.atlassian.net/browse/CORE-2843

rpc/connection.go
maxtaco: You can file a separate ticket for it.

maxtaco: Related to logging, since one of those parameters is for logging.

rpc/user_server.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	if len(errMsgs) == 0 {
+		a.retCh <- nil
+	}
+	a.retCh <- errors.New(strings.Join(errMsgs, ", "))
+
+	if len(s.conns) == 0 {
+		s.tryShutdownCh <- true
```
maxtaco: Nice, also do this eagerly as you hear connections are closing from the RPC client.

rpc/server_test.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	log.Printf("closing conn")
+	conn.Close()
+
+	// server doesn't know the connection is closed yet:
+	s.statsCh <- ch
+	stats = <-ch
+	if stats.UserServerCount != 1 {
+		t.Errorf("user servers: %d, expected 1", stats.UserServerCount)
```
maxtaco: Can know this failure case eagerly.

rpc/user_server.go
```diff
@@ -0,0 +1,117 @@
...
+				conn.close()
+				delete(s.conns, id)
+			}
+		}
+	}
+
+	if len(errMsgs) == 0 {
+		a.retCh <- nil
+	}
+	a.retCh <- errors.New(strings.Join(errMsgs, ", "))
```
maxtaco: two sends to retCh here?

patrickxb:
PTAL @maxtaco 

Server hash of uid to last conn id implemented with perUID confirm before shutdown.

rpc/user_server.go
```diff
@@ -0,0 +1,168 @@
...
+
+			if s.isConnDown(err) {
+				s.removeConnection(conn, id)
+			}
+		}
+	}
+
+	if len(errMsgs) == 0 {
+		a.retCh <- nil
+		return
```
maxtaco: is it bad to short-circuit the `tryShutdownCh` in this case?

rpc/user_server.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// confirm with the server that it is ok to shutdown
+	ok := make(chan bool)
+	args := confirmUIDShutdownArgs{
+		uid:        s.uid,
+		lastConnID: s.lastConnID,
+		ok:         ok,
+	}
+	s.parentConfirmCh <- args
+	if <-ok == false {
```
maxtaco: werid-looking, is this go-y?

patrickxb: I thought `if !<-ok` was worse, but i'll make it two lines.

rpc/user_server.go
patrickxb: good point.

maxtaco:
:ship:

PR #17, opened 2016-03-31
Use ctime of dismissal as dtime for items

patrickxb:
r? @maxtaco 

I haven't come up with a clean way to test this change.  The existing tests pass, but if you have any ideas, please let me know.


protocol/avdl/common.avdl
```diff
@@ -38,6 +38,7 @@ protocol common {
 	record Dismissal {
 		array<MsgID> msgIDs;
 		array<MsgRange> ranges;
+		Time ctime;
```
maxtaco: This shouldn't be necessary. The dismissal always comes with a Metadata, which should have a CTime in it.

storage/schema.go
```diff
@@ -42,6 +42,7 @@ var schema = []string{
 		uid   CHAR(16) NOT NULL,
 		msgid CHAR(16) NOT NULL,
 		dmsgid CHAR(16) NOT NULL, -- "the message IDs to dismiss",
+		ctime DATETIME(6) NOT NULL,
```
maxtaco: this also shouldn't be necessary (following the above comment)

patrickxb:
@maxtaco PTAL

storage/schema.go
```diff
@@ -51,6 +51,7 @@ var schema = []string{
 		msgid CHAR(16) NOT NULL,
 		category VARCHAR(128) NOT NULL,
 		dtime DATETIME(6) NOT NULL, -- "throw out matching events before dtime",
+		ctime DATETIME(6) NOT NULL,
```
maxtaco: can you nuke this?

storage/sql_sm.go
```diff
@@ -127,29 +133,36 @@ func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) err
...
-	qb.Build("WHERE uid=? AND msgid=?")
-	upd, err := tx.Prepare(qb.Query())
+	upd, err := tx.Prepare("UPDATE items SET dtime=? WHERE uid=? AND msgid=?")
 	if err != nil {
 		return err
 	}
 	defer upd.Close()
+
+	var ct time.Time
+	if ctime != nil && ctime.Time() != nil {
```
maxtaco: does this case ever happen?

storage/sql_sm.go
```diff
@@ -127,29 +133,36 @@ func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) err
...
+		ct = *ctime.Time()
+	} else {
+		ct = s.clock.Now()
```
maxtaco: I've intentionally avoided this so that we can use the MySQL's server clock.

storage/sql_sm.go
maxtaco: I would expected `ctime` to always be non-nil here.

storage/sql_sm.go
```diff
@@ -226,32 +258,37 @@ func (s *SQLEngine) consumeInBandMessage(m gregor.InBandMessage) error {
...
+		if err != nil {
+			tx.Rollback()
+		} else {
+			err = tx.Commit()
+		}
+	}()
+
 	md := m.Metadata()
-	if err := s.consumeInBandMessageMetadata(tx, md, gregor.InBandMsgTypeUpdate); err != nil {
+	if err = s.consumeInBandMessageMetadata(tx, md, gregor.InBandMsgTypeUpdate); err != nil {
```
maxtaco: I thought you'd be setting ctime here if it wasn't already set.

maxtaco: (Inside `consumeInBandMessageMetadata`)

storage/sql_sm.go
patrickxb: I was just putting something in if it is empty, which seems to happen in the tests.

storage/sql_sm.go
```diff
@@ -157,21 +170,40 @@ func (s *SQLEngine) consumeMsgIDsToDismiss(tx *sql.Tx, u gregor.UID, mid gregor.
...
 		qb.TimeOrOffset(mr.EndTime())
+		qb.Build(",")
+		qb.Now()
 		qb.Build(")")
 		if err := qb.Exec(tx); err != nil {
 			return err
 		}
 
+		// get ctime from the dismissal row
+		ctime, err := s.ctimeFromDismissalsByTime(tx, u, mid, mr)
```
maxtaco: I'm also not certain why this is needed. Can't `ctime` be passed in as in the corresponding calls to `consumeMsgIDsToDismiss`?

storage/sql_sm.go
patrickxb: (now that it is getting it from metadata)

maxtaco: You should ensure that `ctime` is always set in the metadata first (see comment in this PR somewhere)

storage/schema.go
patrickxb: no, it's needed to let mysql set the time on the insert.

maxtaco: It shouldn't be, `ctime` should be set in the metadata row, not this row.

storage/sql_sm.go
patrickxb: yes.

from our slack discussion yesterday, I was under the impression that the insert into dismissals_* needed to create the ctime, then select from it.  obviously I missed something.

maxtaco: right, but i was trying to say that all of this was in the `consumeMetadata` function.

patrickxb:
Ok, let's try again...PTAL

storage/sql_sm.go
```diff
@@ -127,29 +138,29 @@ func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) err
...
-	qb.Build("WHERE uid=? AND msgid=?")
-	upd, err := tx.Prepare(qb.Query())
+	upd, err := tx.Prepare("UPDATE items SET dtime=? WHERE uid=? AND msgid=?")
 	if err != nil {
 		return err
 	}
 	defer upd.Close()
+
+	qb := s.newQueryBuilder()
+	ctimeArg := qb.TimeArg(ctime)
```
maxtaco: collapse into one line?

storage/sql_sm.go
```diff
@@ -127,29 +138,29 @@ func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) err
...
+	hexUID := hexEnc(u)
+	hexMID := hexEnc(mid)
+
 	for _, dmid := range dmids {
-		_, err = ins.Exec(hexEnc(u), hexEnc(mid), hexEnc(dmid))
+		_, err = ins.Exec(hexUID, hexMID, hexEnc(dmid))
 		if err != nil {
 			return err
 		}
-		qbc := qb.Clone()
```
maxtaco: can get rid of this function now

storage/sql_sm.go
```diff
@@ -203,9 +220,28 @@ func (s *SQLEngine) consumeInBandMessageMetadata(tx *sql.Tx, md gregor.Metadata,
 	qb := s.newQueryBuilder()
 	qb.Build("INSERT INTO messages(uid, msgid, mtype, devid, ctime) VALUES(?, ?, ?, ?,",
 		hexEnc(md.UID()), hexEnc(md.MsgID()), int(t), hexEncOrNull(md.DeviceID()))
-	qb.Now()
+	if md.CTime().IsZero() {
```
maxtaco: will this ever be 0?

maxtaco: or course it will be, i missed the collapse point above. sorry!

test/objects.go
```diff
@@ -219,6 +221,10 @@ func (t *testDismissal) RangesToDismiss() []gregor.MsgRange {
 	return ret
 }
 
+func (t *testDismissal) CTime() time.Time {
```
maxtaco: can we nuke this?

maxtaco:
:+1:

PR #19, opened 2016-04-01
Marshal and store State on Clients

jacobhaven:


protocol/go/extras.go
```diff
@@ -61,18 +62,18 @@ func (d Dismissal) MsgIDsToDismiss() []gregor.MsgID {
 }
 
 type ItemAndMetadata struct {
-	md *Metadata
-	i  *Item
+	MD   *Metadata `codec:"md" json:"md"`
```
maxtaco: if you're doing this, can maybe have the AVDL compiler output it:

```
record {
    union<null, Metadata> md;
    union<null, Item> item;
}
```

A litle bit weird, but it will generate the same code...

vendor/github.com/cenkalti/backoff/backoff.go
```diff
@@ -0,0 +1,59 @@
+// Package backoff implements backoff algorithms for retrying operations.
```
maxtaco: We can't vendor here because gregor is a library that we're going to link with `keybase/client/go`.

client/client.go
```diff
@@ -0,0 +1,55 @@
+package client
```
maxtaco: why does this code live in client/ rather than storage/ ?

client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+
+	b, err := state.Marshal()
+	if err != nil {
+		return err
+	}
+
+	return c.storage.Store(c.user, b)
```
maxtaco: Can we compute the `ctime` of the most recent message easily on restore?  Because when a client comes back up, it will need to request `InBandMessagesSince()` for some time, which should probably be the max `ctime` previously stored.

storage/sql_sm.go
```diff
@@ -426,6 +426,45 @@ func (s *SQLEngine) rowToInBandMessage(u gregor.UID, rows *sql.Rows) (gregor.InB
...
+	for _, i := range items {
+		u := i.Metadata().UID()
+		if !s.rowExists(tx, u, i.Metadata().MsgID()) {
+			if err := s.consumeCreation(tx, u, i); err != nil {
+				return err
+			}
+		}
+	}
+
+	return tx.Commit()
```
maxtaco: you should `Rollback` in the case of a failure.  Probably want to do so in a `defer`, for if you quit early

storage/sql_sm.go
```diff
@@ -426,6 +426,45 @@ func (s *SQLEngine) rowToInBandMessage(u gregor.UID, rows *sql.Rows) (gregor.InB
...
+	if err != nil {
+		return err
+	}
+
+	tx, err := s.driver.Begin()
+	if err != nil {
+		return err
+	}
+
+	upd, err := tx.Prepare("UPDATE items SET dtime=? WHERE uid=? AND msgid=?")
```
maxtaco: do you need this query?  it doesn't seem like you ever execute it.

jacobhaven: I removed that logic, but actually, I guess it is needed in the case where the new item's `dtime` is set and the the one in the DB is `null`.

client/client.go
jacobhaven: Sure, I think is makes sense to have `Client#Restore() (gregor.TimeOrOffset, error)` returning the most recent  `ctime` that wasn't already in the `StateMachine`.

vendor/github.com/cenkalti/backoff/backoff.go
jacobhaven: Alright... There was already code vendored in master, I just updated it. Should I remove the `vendor` directory entirely?

client/client.go
jacobhaven: Or actually, we could just treat `InBandMessagesSince(u, d, nil)` as "Return all `InBandMessage`s since the latest `CTime`" (this would naively require iterating through all the items, but we could structure the `StateMachine`s to make it a simpler operation to get the latest `CTime`)

jacobhaven: I'll save implementing this logic for another PR.

jacobhaven:
@patrickxb @maxtaco PR comments addressed and rebased off master. Please take a look

storage/sql_sm.go
```diff
@@ -426,6 +435,31 @@ func (s *SQLEngine) rowToInBandMessage(u gregor.UID, rows *sql.Rows) (gregor.InB
...
+	items, err := state.Items()
+	if err != nil {
+		return err
+	}
+
+	tx, err := s.driver.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
```
patrickxb: This means rollback will be called regardless of error status.  Perhaps that's ok after a commit?  The pattern I'm used to is commit if no errors, rollback if there is an error.

jacobhaven: Yes, `tx.Rollback()` after `tx.Commit()` is a noop. (commit sets a `done` flag that rollback checks before doing anything)

storage/sql_sm.go
```diff
@@ -99,7 +99,7 @@ func (s *SQLEngine) newQueryBuilder() *queryBuilder {
 func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) error {
 	md := i.Metadata()
 	qb := s.newQueryBuilder()
-	qb.Build("INSERT INTO items(uid, msgid, category, body, dtime) VALUES(?,?,?,?,",
+	qb.Build(`REPLACE INTO items(uid, msgid, category, body, dtime) VALUES(?,?,?,?,`,
```
maxtaco: remind me which cases you saw that required this change?

maxtaco: AddState?

storage/client_test.go
```diff
@@ -0,0 +1,44 @@
+package storage
```
maxtaco: Do you think it would work to call `TestStateMachineAllDevices` on your new Client StateMachine store?

storage/client.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+
+	state, err := c.objFactory.UnmarshalState(value)
+	if err != nil {
+		return err
+	}
+
+	return c.sm.AddState(state)
```
maxtaco: What's the use case for when `Restore()` would do anything other than load the whole state from storage?  In other words, why bother to perform this unioning?

jacobhaven: If the client state is a `SQLEngine` then it will already have state that needs to be merged with the restored state.

Side-benefit: this also allows a `MemEngine` to receive notifications before/during its restore, rather than blocking on what could potentially be a long process.

storage/sql_sm.go
jacobhaven: Yes. It's simpler to write it this way instead of `UPDATE`ing `dtime`.

storage/client.go
maxtaco: But my plan was to call `InBandMessagesSince` in that case....

storage/client_test.go
jacobhaven: It required rewriting all the tests helpers (CORE-2808), but *done*

interface.go
```diff
@@ -151,6 +164,7 @@ type ObjFactory interface {
 	MakeState(i []Item) (State, error)
 	MakeMetadata(uid UID, msgid MsgID, devid DeviceID, ctime time.Time, i InBandMsgType) (Metadata, error)
 	MakeInBandMessageFromItem(i Item) (InBandMessage, error)
+	UnmarshalState([]byte) (State, error)
```
maxtaco: Why do we need both the `UnmarshalState` method here and the `InitState` method above?

protocol/go/factory.go
```diff
@@ -37,20 +44,16 @@ func castDeviceID(d gregor.DeviceID) (DeviceID, error) {
...
-func timeToTimeOrOffset(timeIn *time.Time) TimeOrOffset {
-	var timeOut Time
-	if timeIn != nil && !timeIn.IsZero() {
-		timeOut = ToTime(*timeIn)
-	}
-	return TimeOrOffset{
-		Offset_: 0,
-		Time_:   timeOut,
+func timeToTimeOrOffset(timeIn *time.Time) (too TimeOrOffset) {
+	if timeIn != nil {
```
maxtaco: Why is it OK to drop the `IsZero` check?

protocol/go/factory.go
```diff
@@ -163,10 +166,20 @@ func (o ObjFactory) MakeInBandMessageFromItem(i gregor.Item) (gregor.InBandMessa
...
-			Creation_: ourItem.i,
+			Md_:       *ourItem.Md_,
+			Creation_: ourItem.Item_,
 		},
 	}, nil
 }
 
+func (o ObjFactory) UnmarshalState(b []byte) (gregor.State, error) {
+	var items []ItemAndMetadata
+	err := codec.NewDecoderBytes(b, new(codec.MsgpackHandle)).Decode(&items)
```
maxtaco: set `WriteExt = true` on the code.

protocol/go/extras.go
```diff
@@ -179,8 +181,14 @@ func (s State) Items() ([]gregor.Item, error) {
 	return ret, nil
 }
 
+func (s State) Marshal() ([]byte, error) {
+	var b []byte
+	err := codec.NewEncoderBytes(&b, new(codec.MsgpackHandle)).Encode(s.items)
```
maxtaco: set `WriteExt = true` on the codec handle

storage/mem_sm.go
```diff
@@ -315,3 +325,40 @@ func (m *MemEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t grego
...
+	}
+
+	for u, items := range userItems {
+		u.addItems(items)
+	}
+
+	return nil
+}
+
+func (m *MemEngine) RemoveDismissed(too gregor.TimeOrOffset) {
```
maxtaco: why are you ignoring the passed in `TimeOrOffset`?  The intention is that you can prune out some of the dismissed itmes but only if their dismissal time was more than a given time away....

test/helpers.go
```diff
@@ -0,0 +1,312 @@
+package test
+
+import (
+	"crypto/rand"
+	"testing"
+	"time"
+
+	"github.com/jonboulle/clockwork"
+	gregor "github.com/keybase/gregor"
+	protocol "github.com/keybase/gregor/protocol/go"
```
maxtaco: My original hope was for these test helpers to not depend on `protocol`.  That is, we can pass an `ObjFactory` in to the exported test routines.  It seems like our `ObjFactory` interface would need a few more methods to do so, but it seems very doable... What do you think?

maxtaco:
I have a fix for that protocol dependency in my branch, but I'm noticing some test failures. There's the C/I failure above (which I haven't seen locally) and there's also this that shows up half the time:

```
--- FAIL: TestLevelDBClient (0.00s)
        Error Trace:    helpers.go:38
			helpers.go:306
			helpers.go:308
	Error:  	Not equal: []string{"f2", "f3"} (expected)
			        != []string{"f3", "f2"} (actual)
			
			Diff:
			--- Expected
			+++ Actual
			@@ -1,4 +1,4 @@
			 ([]string) (len=2 cap=2) {
			- (string) (len=2) "f2",
			- (string) (len=2) "f3"
			+ (string) (len=2) "f3",
			+ (string) (len=2) "f2"
			 }
	Messages:	the right values in the store		
```

I've seen this bug before in other engines.  It means that engine is inserting these items out of order.

maxtaco:
Another failure I noticed: doesn't work with MySQL anymore:

```
--- FAIL: TestMySQLEngine (0.12s)
	helpers.go:132: In inserting msg m1: Error 1406: Data too long for column 'uid' at row 1
```

maxtaco:
As I was typing up the commit message above, I figured out why sometimes mem and client tests are failing.  I also remembered why I initially stored the items in a dump list, and not in a map (as you changed it to).

The problem is that now the messages aren't ordered by ctime, but the tests want them to be. The tests make sure that items in the same category show up in the right order.  Probably the easiest solution is just to sort by `Ctime` on the way out of the `State()` call.  So in other words, you only have to make 1 additional struct, and 4 additional methods :wink:

protocol/go/factory.go
jacobhaven: If `timeIn` is zero, then `timeOut` also gets left as zero.

storage/mem_sm.go
jacobhaven: Ah, this logic got removed with the `TimeOrOffset#Before(time.Time)` method. I may have to add that back.

This will only work with exact times, not offsets, is that the intention?

protocol/go/factory.go
maxtaco: Ah, you're right, because `ToTime` does an explicit zero check. I was worried about the bug mentioned in `ToZero`, but we're covered.

storage/mem_sm.go
jacobhaven: Actually, given that `State(UID,DeviceID, t TimeOrOffset)` already does [this pruning of dismissed messages](https://github.com/keybase/gregor/pull/19/files/8951cc41f1f912bcdbee7e19a196ed4d2c907243#diff-82a6542504da5b983212cdbfa074e098R192), I'm not sure this method is even necessary. The dismissed messages before `t` will disappear from the `MemEngine` automatically when it's serialized.

storage/sql_sm.go
```diff
@@ -82,7 +83,7 @@ func (q *queryBuilder) Exec(tx *sql.Tx) error {
 func hexEnc(b byter) string { return hex.EncodeToString(b.Bytes()) }
 
 func hexEncOrNull(b byter) interface{} {
-	if b == nil {
+	if b == nil || len(b.Bytes()) == 0 {
```
jacobhaven: Maybe we should have a
```Go
func isEmpty(b byter) bool {
	return b == nil || len(b.Bytes()) == 0 
}
```
but I think this is the only place that check is necessary.

storage/mem_sm.go
maxtaco: :+1: good point

storage/sql_sm.go
maxtaco: sure, this would be useful when comparing `DeviceID`s somewhere in this PR.  Are you hitting a case in which IDs are coming back as either empty arrays or `nil`?

test/helpers.go
jacobhaven: Seems reasonable, but I might push that to a new PR.

maxtaco: Sounds good.  You might also need some of the bugfixes in `protocol/` to get C/I happy about your changeset though.

storage/sql_sm.go
jacobhaven: The `gregor1` protocol package always returns empty slices, not `nil`.

storage/mem_sm.go
```diff
@@ -56,13 +56,13 @@ type loggedMsg struct {
 // user consists of a list of items (some of which might be dismissed) and
 // and an unpruned log of all incoming messages.
 type user struct {
-	items [](*item)
+	items map[string](*item)
```
jacobhaven: This change breaks consistent ordering of items. I'm going to revert it.

jacobhaven:
I've fixed all these issues (except the MySQL "Data too long for column 'uid'" coming from a double-hex conversion that @maxtaco has fixed on another branch).

storage/sql_schema.go
```diff
@@ -9,17 +9,17 @@ var schema = []string{
 	`DROP TABLE IF EXISTS messages`,
 
 	`CREATE TABLE messages (
-		uid   CHAR(16) NOT NULL,
-		msgid CHAR(16) NOT NULL,
+		uid   CHAR(32) NOT NULL,
```
jacobhaven: Shouldn't 16 bytes still be correct here. The problem was `MakeUID`/`MakeDeviceID` was taking 8 bytes, hex encoding them to 16 byes, and then `SQLEngine` was hex encoding them *again* to 32 bytes.

Removing the hex encoding in `MakeUID`/`MakeDeviceID` should solve this issue by itself.

interface.go
```diff
@@ -148,6 +149,9 @@ type StateMachine interface {
 
 	// ObjFactory returns the ObjFactory used by this StateMachine.
 	ObjFactory() ObjFactory
+
+	// Clock returns the clockwork.Clock used by this StateMachine.
+	Clock() clockwork.Clock
```
jacobhaven: We don't _need_ this, but it does simplify maintaining the same clock associated with individual `StateMachine`. All of our engines have their own internal `clockwork.Clock`, so exposing that through the interface isn't a bad idea.

jacobhaven:
Alright, this LGTM. @maxtaco or @patrickxb any more comments?

interface.go
```diff
@@ -131,11 +133,25 @@ type StateMachine interface {
...
 	// for the user u on device d.  If d is nil, then we'll return
 	// all messages across all devices.  If d is a device, then we'll
 	// return global messages and per-device messages for that device.
 	InBandMessagesSince(u UID, d DeviceID, t TimeOrOffset) ([]InBandMessage, error)
+
+	// ObjFactory returns the ObjFactory used by this StateMachine.
+	ObjFactory() ObjFactory
+
+	// Clock returns the clockwork.Clock used by this StateMachine.
+	Clock() clockwork.Clock
```
jacobhaven: We don't _need_ this, but it does simplify maintaining the same clock associated with individual `StateMachine`. All of our engines have their own internal `clockwork.Clock`, so exposing that through the interface isn't a bad idea.

maxtaco: I'm good with this.

maxtaco:
One more thing: you said the protocol layer can make empty byte arrays for `nil` UIDs.  Can we change this so they can just return `nil`?

protocol/go/factory.go
```diff
@@ -1,35 +1,38 @@
...
 }
 
 func castDeviceID(d gregor.DeviceID) (DeviceID, error) {
+	if d == nil {
+		return DeviceID(nil), nil
+	}
 	ret, ok := d.(DeviceID)
 	if !ok {
-		return DeviceID(""), fmt.Errorf("Bad Device ID; wrong type")
+		return DeviceID(""), errors.New("bad Device ID; wrong type")
```
maxtaco: `nil` rather than `""`?

protocol/go/factory.go
```diff
@@ -1,35 +1,38 @@
...
 func (o ObjFactory) MakeCategory(s string) (gregor.Category, error) { return Category(s), nil }
 
 func castUID(uid gregor.UID) (UID, error) {
+	if uid == nil {
+		return UID(nil), nil
+	}
 	ret, ok := uid.(UID)
 	if !ok {
-		return UID(""), fmt.Errorf("Bad UID; wrong type")
+		return UID(""), errors.New("bad UID; wrong type")
```
maxtaco: `nil` rather than `""`?

jacobhaven:
Well, they are returning `[]byte(nil)`. I think the easiest thing is to always check `b != nil && b.Bytes() != nil`

protocol/go/factory.go
jacobhaven: I updated these, but it doesn't matter (especially because it's returning an error).

storage/mem_sm.go
```diff
@@ -108,9 +108,9 @@ func msgIDtoString(m gregor.MsgID) string {
 }
 
 func (u *user) dismissMsgIDs(now time.Time, ids []gregor.MsgID) {
-	set := make(map[string]struct{})
+	set := make(map[string]bool)
```
maxtaco: any reason not to use `struct{}`s here btw?

maxtaco:
:+1: (after one last q about `map[string]bool` versus `map[string]struct{}`)

storage/mem_sm.go
jacobhaven: `bool` is more idiomatic (and shorter). I don't think there's a real reason to use an empty `struct{}{}` sentinel instead.

maxtaco: I'd seen it both ways!  @jzila / @patrickxb / @akalin-keybase quick straw poll: `map[string]bool` or `map[string]struct{}` ?

strib: `bool` 4eva: https://blog.golang.org/go-maps-in-action

maxtaco: @jzila says `map[string]bool` FWIW

maxtaco:
:+1:

storage/mem_sm.go
patrickxb: bool

akalin-keybase: boooool

jzila: to be fair I got that from @akalin-keybase months ago :)

akalin-keybase: I also got it from @strib months ago (citing the same blog post).

maxtaco: Why did I ever think `map[string]struct{}` was the right idea then?

PR #20, opened 2016-04-01
Remove vendoring, add dummy authentication to gregord

patrickxb:
This adds dummy authentication to gregord.

Vendoring causing issues once keybase/go/client integration performed, so I removed it all.

Tried to vendor gregord only, but that also had issues.

So in the interest of moving forward, I just got rid of all the vendoring.

r? @maxtaco


maxtaco:
@patrickxb  what's up with the C/I failure?

maxtaco:
:+1:

maxtaco:
(Is this project now mature enough that we have our first flakey C/I test?)

patrickxb:
I don't know...the only thing I can think is that it has something to do with a vendored package?  Of course there's no way to tell which version travis is using...presumably the latest, but...

maxtaco:
My money is on a race condition in the tests that only shows up when running on a slow system like CI. Your theory wouldn't explain why sometimes it works, and others not.  Well, let's scrutinize the logs the next time there's a failure.

patrickxb:
Well there were some issues like this with old msgpack when I was developing.

maxtaco:
I think travis will fetch master, so if that's what you vendored, it shouldn't matter.

PR #21, opened 2016-04-05
RPC Integration with Session Server

jacobhaven:


gregord/config.go
```diff
@@ -14,10 +12,13 @@ import (
...
 	"strings"
 	"time"
+
+	"github.com/goamz/goamz/aws"
+	"github.com/goamz/goamz/s3"
 )
 
 type Options struct {
-	SessionServer *url.URL
+	SessionServer net.Addr
```
maxtaco: elsewhere we've adopted the convention:

```
fmprpc://foo.bar.com:333
fmprpc+tls://bar.bam.bop:443
```

CC: @patrickxb

gregord/main.go
```diff
@@ -1,31 +1,27 @@
...
+	srv := grpc.NewServer()
+	conn, err := net.Dial(opts.SessionServer.Network(), opts.SessionServer.String())
 	if err != nil {
-		errorf("%s\n", err)
-		os.Exit(2)
+		log.Fatal(err)
 	}
-	return
+
+	srv.SetAuthenticator(grpc.NewAuthClient(rpc.NewClient(rpc.NewTransport(conn, nil, nil), nil)))
```
maxtaco: todo to:
* pass in appropriate logger
* handle correct error wrapping / unwrapping.  should errors just be strings, or should they be msgpack structs with status codes and descriptions.  Not decided yet but I'm thinking we should follow the same convention we use everywhere else

maxtaco: What happened to `newMainServer...listenAndServe...`?

rpc/auth.go
```diff
@@ -1,11 +1,30 @@
 package rpc
 
 import (
+	keybase1 "github.com/keybase/client/go/protocol"
```
maxtaco: can we move anything that depends on `keybase` into a separate subdir?

rpc/auth.go
```diff
@@ -1,11 +1,30 @@
...
-	Authenticate(ctx context.Context, tok protocol.AuthToken) (protocol.UID, protocol.SessionID, error)
+	Authenticate(ctx context.Context, tok string) (protocol.UID, string, error)
+}
+
+type AuthClient struct {
+	keybase1.QuotaClient
+}
+
+func NewAuthClient(Cli rpc.GenericClient) *AuthClient {
+	return &AuthClient{keybase1.QuotaClient{Cli}}
```
maxtaco: Let's not use `QuotaClient`, since I want to deprecate that.  Can you make a new protocol for sessiond?  It doesn't have to live in the `keybase/client` repo, it can actually live in this repo.  It's basically:

```
@namespace("gregor.1")
protocol session {
 
  record AuthenticateSessionRes {
    string uid; // encoded in Hex; we can skip the keybase dependency by just having this a string
    string sid;
    int generated;
    int lifetime; 
  }

  AuthenticateSessionRes authenticateSession(string session);
}
```

gregord/main.go
maxtaco: Actually, the session server is currently using the error convention of passing structs.

See [this code](https://github.com/keybase/client/blob/66e321ff701457eccf04048a18b624ad11ee6576/go/libkb/rpc_exim.go#L176-L197) for where that happens in the service and kbfsfuse.

maxtaco:
Thanks for the early check-in, I appreciate it.

Some todos and suggestions noted.  Also, we should start thinking about how to test all of this, both in unit and integration.

jacobhaven:
@maxtaco Yeah, unit testing was my main concern going forward with this. We'll need to make a dummy rpc auth server that we spawn and use for unit tests.

gregord/config.go
jacobhaven: So, then do we do string matching on the `URL#Scheme` and open a TLS Conn if it's `fmprpc+tls`?

maxtaco: I think so -- @patrickxb is that right?

patrickxb: Yes, see these files:

https://github.com/keybase/client/blob/master/go/service/rpc.go
https://github.com/keybase/client/blob/master/go/service/rpc_test.go

protocol/gregor1/avdl/auth.avdl
```diff
@@ -1,6 +1,12 @@
 @namespace("gregor.1")
 
 protocol auth {
-	UID authenticateSessionToken(string tok);
-	void revokeSessionToken(string tok);
+	@typedef("string") record SessionID {}
```
maxtaco: i can make this `[]byte` if you prefer

jacobhaven: A slice can't be a key in a map... So, `string` is easier to work with.

maxtaco: :+1:

maxtaco:
@jacobhaven is this WIP or ready for review?

jacobhaven:
@maxtaco Should be ready for review. It probably needs a more comprehensive mock of the sessiond server and possibly the (small) addition of Gregor client msgpack-rpc code (though that's probably better in a separate PR for CORE-2706)

maxtaco:
ok, i'll take look tomorrow, thanks @jacobhaven!

gregord/config.go
```diff
@@ -52,6 +53,67 @@ Environment Variables
     -s3-config-bucket or S3_CONFIG_BUCKET
 `
 
+type fmpURIScheme string
```
maxtaco: I'm thinking the right place for this is in the keybase/go-framed-msgpack-rpc library.  @jzila / @akalin-keybase what do you think?

maxtaco: And BTW @strib this is probably the right solution to the problem we were hitting yesterday...

gregord/config.go
```diff
@@ -52,6 +53,67 @@ Environment Variables
...
+
+	f := &fmpURI{HostPort: uri.Host}
+
+	switch fmpURIScheme(uri.Scheme) {
+	case fmpSchemeStandard:
+		f.Scheme = fmpSchemeStandard
+	case fmpSchemeTLS:
+		f.Scheme = fmpSchemeTLS
+	default:
+		return nil, fmt.Errorf("invalid framed msgpack rpc scheme %s", uri.Scheme)
```
maxtaco: can you test this case in the usage tests below?

rpc/session.go
```diff
@@ -0,0 +1,48 @@
...
+		sessions:   make(map[gregor1.SessionToken]gregor1.AuthResult),
+		sessionIDs: make(map[gregor1.SessionID]gregor1.SessionToken),
+	}
+}
+
+func (sc *sessionCacher) AuthenticateSessionToken(ctx context.Context, tok gregor1.SessionToken) (res gregor1.AuthResult, err error) {
+	var ok bool
+	if res, ok = sc.sessions[tok]; ok {
+		return
+	}
```
maxtaco: can you make sessions expire after about ~10 minutes?

rpc/server_test.go
```diff
@@ -9,25 +9,52 @@ import (
...
+type mockAuth struct {
+	sessions   map[gregor1.SessionToken]gregor1.AuthResult
+	sessionIDs map[gregor1.SessionID]gregor1.SessionToken
+}
+
+func (m mockAuth) AuthenticateSessionToken(_ context.Context, tok gregor1.SessionToken) (gregor1.AuthResult, error) {
+	if res, ok := m.sessions[tok]; ok {
+		return res, nil
+	}
+	return gregor1.AuthResult{}, errors.New("invalid token")
```
maxtaco: is there a unit test to test this case?

maxtaco:
Is session invalidation fully plumbed?  It seems like there's no way to take a connection out of the "authenticated" state. What's the plan for that?

gregord/config.go
jacobhaven: I can just copy the tests from `keybase/client`, or better: move both to `keybase/go-framed-msgpack-rpc`

maxtaco: No, that's not what I meant.  See below, we test wonky parameters being passed in.  Make a single test for a wonky URL here to pick up the case I mentioned

rpc/server_test.go
jacobhaven: Yes https://github.com/keybase/gregor/blob/jacob/CORE-2707/rpc/server_test.go#L151

gregord/flag_test.go
```diff
@@ -27,19 +28,19 @@ func TestUsage(t *testing.T) {
 	testBadUsage(t, []string{"gregor", "--bind-address", "localhost:65537", "--session-server", "localhost"}, ebu, "bad port (\"65537\") in bind-address")
 	testBadUsage(t, []string{"gregor", "--bind-address", "localhost:-20", "--session-server", "localhost"}, ebu, "bad port (\"-20\") in bind-address")
 	testBadUsage(t, []string{"gregor", "--bind-address", ":4000"}, ebu, "No session-server URI specified")
-	testBadUsage(t, []string{"gregor", "--bind-address", ":4000", "--session-server", "localhost", "--tls-key", "hi"},
+	testBadUsage(t, []string{"gregor", "--bind-address", ":4000", "--session-server", "fmprpc://localhost:30000", "--tls-key", "hi"},
 		ebu, "you must provide a TLS Key and a TLS cert, or neither")
-	testBadUsage(t, []string{"gregor", "--bind-address", ":4000", "--session-server", "localhost", "--tls-cert", "hi"},
+	testBadUsage(t, []string{"gregor", "--bind-address", ":4000", "--session-server", "fmprpc://localhost:30000", "--tls-cert", "hi"},
```
maxtaco: insert code, or some such...

```go
testBadUsage(t, []string{"gregor", "--bind-address", ":4000", "--session-server", "XXXXX://localhost:30000", "--tls-cert", "hi"}
```

gregord/config.go
akalin-keybase: Sure. Will the rpc library eventually use this constant?

maxtaco: My thought was that we should make a URI format up for fmprpc for command-line input of URIs. We already have this code in two places after this commit -- @patrickxb  where did you recently put in this code?

patrickxb: You mean this?

https://github.com/keybase/client/blob/master/go/service/rpc.go
https://github.com/keybase/client/blob/master/go/service/rpc_test.go

maxtaco: y

patrickxb: (used here:

https://github.com/keybase/client/blob/master/go/service/main.go#L236

rpc/session.go
```diff
@@ -1,46 +1,135 @@
...
+func (sc *sessionCacher) requestHandler() {
+	expiryCh := make(chan gregor1.SessionID)
+	var expiryQueue []expirySt
+	for {
+		select {
+		case sid := <-expiryCh:
+			sc.deleteSID(sid)
+			expiryQueue = expiryQueue[1:]
+			if len(expiryQueue) > 0 {
+				go expireAtTime(expiryCh, expiryQueue[0])
```
strib: Instead of a goroutine+Sleep here and below, you can just do  `expiryCh = time.After(e.expiry.Sub(time.Now()))`.  Hopefully it does the right thing in the case that the duration is negative, but I can't find documentation about it.

strib: (And you can just get the `sid` from `expiryQueue[0].sid`.)

rpc/session.go
```diff
@@ -1,46 +1,135 @@
...
+// NewSessionCacher creates a new AuthInterface that caches sessions for the given timeout.
+func NewSessionCacher(a gregor1.AuthInterface, timeout time.Duration) gregor1.AuthInterface {
+	sc := &sessionCacher{
 		a:          a,
+		timeout:    timeout,
 		sessions:   make(map[gregor1.SessionToken]gregor1.AuthResult),
 		sessionIDs: make(map[gregor1.SessionID]gregor1.SessionToken),
+		reqCh:      make(chan request),
+	}
+	go sc.requestHandler()
```
strib: Not sure how Gregor tests work, but usually you'd want a way to shut down this goroutine at the end of each test.

rpc/session.go
```diff
@@ -1,46 +1,135 @@
...
+	for {
+		select {
+		case sid := <-expiryCh:
+			sc.deleteSID(sid)
+			expiryQueue = expiryQueue[1:]
+			if len(expiryQueue) > 0 {
+				go expireAtTime(expiryCh, expiryQueue[0])
+			}
+		case req := <-sc.reqCh:
+			switch req := req.(type) {
```
strib: default case that logs an error?

rpc/session.go
```diff
@@ -1,46 +1,135 @@
...
+		}
 	}
 }
 
 func (sc *sessionCacher) AuthenticateSessionToken(ctx context.Context, tok gregor1.SessionToken) (res gregor1.AuthResult, err error) {
-	var ok bool
-	if res, ok = sc.sessions[tok]; ok {
+	respCh := make(chan authResp)
+	sc.reqCh <- readTokReq{tok, respCh}
+	if resp := <-respCh; resp.ok {
```
strib: If `ctx` can ever be canceled, you might want to select on `ctx.Done()` as well to avoid blocking forever.

rpc/session.go
```diff
@@ -1,46 +1,135 @@
...
 
 func (sc *sessionCacher) RevokeSessionIDs(ctx context.Context, sessionIDs []gregor1.SessionID) error {
 	for _, sid := range sessionIDs {
-		if tok, ok := sc.sessionIDs[sid]; ok {
-			if _, ok := sc.sessions[tok]; ok {
-				delete(sc.sessions, sc.sessionIDs[sid])
-			}
-			delete(sc.sessionIDs, sid)
-		}
+		sc.reqCh <- deleteSIDReq{sid}
```
strib: Again, this send can block a long time since `reqCh` is unbuffered.  If `ctx` can be canceled or timeout, it might be good to respect that.

gregord/config.go
jzila: This can be:
```go
switch f.Scheme = fmpURIScheme(uri.Scheme); f.Scheme {
    case fmpSchemeStandard, fmpSchemeTLS:
    default:
        return nil, ...
}
```

gregord/config.go
jzila: I'm not sure why this belongs in RPC. Does RPC need to know about it?

maxtaco: For the same reason connection.go wound up there.  Two consumers of it in two different repos.

jzila: Ok but connection.go is conceptually highly related to what RPC is doing: it allows you to establish a reliable RPC transport. How is the fmpURIScheme related to what RPC does?

maxtaco: often used in concert.

rpc/session.go
jacobhaven: Good point about reading `expiryQueue[0].sid`. I think `time.After` immediately returns on 0/negative duration (`time.Sleep` does). I'll test that.

rpc/session.go
jacobhaven: I initially had a panic there, but that case can never happen (this file is all that sends on `sc.reqCh`)

rpc/session.go
jacobhaven: These requests shouldn't block for any appreciable amount of time., they're just lookups in a map. `sessionCacher` won't even be handling concurrent calls (although it could).

gregord/config.go
jacobhaven: https://github.com/keybase/go-framed-msgpack-rpc/pull/73

gregord/main.go
```diff
@@ -1,31 +1,29 @@
...
+	srv := grpc.NewServer()
+	conn, err := opts.SessionServer.Dial()
 	if err != nil {
-		errorf("%s\n", err)
-		os.Exit(2)
+		log.Fatal(err)
 	}
-	return
+
+	Cli := rpc.NewClient(rpc.NewTransport(conn, nil, nil), nil)
```
jacobhaven: Q: What are the proper rpc `LogFactory` and `WrapErrorFunc`s to use here.

@maxtaco @patrickxb

maxtaco: See what `libkb` does for some suggestions.  The wrap/unwrap error code comes from `rpc_exim.go`

jacobhaven:
> @maxtaco Is session invalidation fully plumbed? It seems like there's no way to take a connection out of the "authenticated" state. What's the plan for that?

I think it makes sense to just recheck the connection auth on every message send/receive. (It should almost always be hitting a local cache). [Latest commit](https://github.com/keybase/gregor/pull/21/commits/18b2eccac29e106ce1fba51d1855f0bcae79b241) does this.

gregord/main.go
jacobhaven: It doesn't seem like error wrap/unwraping is really necessary here.

maxtaco: If you get an error, it will throw a second error since the unmarshall of the erro rwill fail

jacobhaven: Maybe it makes more sense to add an `rpc.defaultErrorUnwrapper` that is used when it's set to `nil`.

maxtaco: There is one, but it's expecting an error object to be just a string.  The API server is sending errors as objects so that we can convert them to and from specific Go types.

gregord/main.go
```diff
@@ -1,31 +1,30 @@
...
+	srv := grpc.NewServer()
+	conn, err := opts.SessionServer.Dial()
 	if err != nil {
-		errorf("%s\n", err)
-		os.Exit(2)
+		log.Fatal(err)
 	}
-	return
+
+	Cli := rpc.NewClient(rpc.NewTransport(conn, nil, nil), libkb.ErrorUnwrapper{})
```
maxtaco: `NewTransport` also needs  `libkb.WrapError` to deal with protocol- and method not founds.

maxtaco: This is OK for now, but let's get rid of this dependency via copy-paste in https://keybase.atlassian.net/browse/CORE-2878

jzila: does that mean needing to reimplement `MethodNotFoundError` and `ProtocolNotFoundError`?

maxtaco: yeah i guess so... hmg.  i don't want to create extra churn by incurring yet another refactor across multiple repos. maybe better to leave as is.

jzila: Well, these refactors are best when they happen organically on-demand, and it sounds like there's demand right now :)

rpc/session.go
```diff
@@ -0,0 +1,152 @@
+package rpc
+
+import (
+	"time"
```
maxtaco: should use `clockwork` so we can test timeout via mocked clock.

rpc/session.go
```diff
@@ -0,0 +1,152 @@
...
+	"golang.org/x/net/context"
+)
+
+// sessionCacher implements gregor1.AuthInterface with a cache of authenticated sessions.
+type sessionCacher struct {
+	a          gregor1.AuthInterface
+	timeout    time.Duration
+	sessions   map[gregor1.SessionToken]gregor1.AuthResult
+	sessionIDs map[gregor1.SessionID]gregor1.SessionToken
+	reqCh      chan request
```
maxtaco: should keep the mocked clock in here rather than the global clock `time` above

rpc/session.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// NewSessionCacher creates a new AuthInterface that caches sessions for the given timeout.
+func NewSessionCacher(a gregor1.AuthInterface, timeout time.Duration) gregor1.AuthInterface {
```
maxtaco: add another param for the mocked clock.

gregord/main.go
maxtaco: It's a nice-to-have, but not as nice as shipping M3!

rpc/session_cacher_test.go
```diff
@@ -45,7 +47,7 @@ func TestSessionCacher(t *testing.T) {
 	checkBad(t, sc, badToken)
 	checkGood(t, sc, goodToken, goodUID)
 
-	// Sleep until timeout, cached results gone
-	time.Sleep(d)
+	// Advance past timeout, cached results gone
+	fc.Advance(d)
```
maxtaco: I'm surprised you don't need 100 + episilon msec here

jacobhaven: Nah, Advancing will instantly trigger a channel receive in the select-loop that removes the token before the next lookup.

maxtaco:
:+1: :ship:

PR #22, opened 2016-04-07
fix the deadlock in per-user ID shutdown

maxtaco:
- simplify the shutdown path
- leave in debug code
- still several race conditions in the test code (see CORE-2882)

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -3,8 +3,10 @@ package rpc
 import (
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"log"
 	"net"
+	"time"
 
 	"github.com/jonboulle/clockwork"
 	gregor "github.com/keybase/gregor"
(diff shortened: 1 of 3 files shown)


maxtaco:
Don't merge this, but I think this is a better approach and doesn't deadlock reliably as the previous iteration did.

rpc/server.go
```diff
@@ -137,27 +141,33 @@ func (s *Server) confirmUIDShutdown(a confirmUIDShutdownArgs) {
...
 		// remove the perUIDServer from users, lastConns
 		delete(s.users, k)
 		delete(s.lastConns, k)
-		a.ok <- true
+
+		// Non-blocking send that it should self-destruct
+		if su != nil {
+			select {
+			case su.shutdownCh <- struct{}{}:
+			default:
```
maxtaco: hmm this isn't exactly right since it won't get send at all in most cases.... maybe we should put a buffer on this channel.   a buffer of size 1 is fine, since this message can only be sent once.

PR #23, opened 2016-04-07
performance fix: broadcasts are async  …

maxtaco:
- sending a broadcast should not block the main thread, so make the whole broadcast pipeline async
- this breaks tests, but they were broken previously (see CORE-2882) so we need to explore a different testing strategy.  in particular, we might need artificial synchronization to wait until clients get broadcasts, available just for testing.  if broadcast is async, there's nothing internally to sychronize on.  but the clients can tell us in the test when the get the message, and/or we can deadlock indefinitely if they never arrive (which will also break tests)

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -3,7 +3,6 @@ package rpc
 import (
 	"encoding/hex"
 	"errors"
-	"fmt"
 	"log"
 	"net"
 
@@ -100,8 +99,6 @@ func (s *Server) getPerUIDServer(u gregor.UID) (*perUIDServer, error) {
 	if err != nil {
 		return nil, err
 	}
-	fmt.Printf("A\n")
-	fmt.Printf("B\n")
 	ret := s.users[k]
 	if ret != nil {
 		return ret, nil
@@ -198,9 +195,7 @@ func (s *Server) BroadcastMessage(c context.Context, m gregor.Message) error {
 }
 
 func (s *Server) sendBroadcast(c context.Context, m protocol.Message) error {
-	fmt.Printf("Q1\n")
 	srv, err := s.getPerUIDServer(gregor.UIDFromMessage(m))
-	fmt.Printf("Q2\n")
 	if err != nil {
 		return err
 	}
@@ -230,13 +225,9 @@ func (s *Server) serve() error {
 		case a := <-s.broadcastCh:
 			s.sendBroadcast(a.c, a.m)
 		case c := <-s.statsCh:
-			fmt.Printf("Z20.1\n")
 			s.reportStats(c)
-			fmt.Printf("Z20.2\n")
 		case a := <-s.confirmCh:
-			fmt.Printf("Z10\n")
 			s.confirmUIDShutdown(a)
-			fmt.Printf("Z10.1\n")
 		case <-s.closeCh:
 			return nil
 		}
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"time"
 
 	"github.com/jonboulle/clockwork"
 	gregor "github.com/keybase/gregor"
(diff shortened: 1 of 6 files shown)


maxtaco:
Don't merge this, this is just a demo of what needs to be done. still has debug code, etc.

rpc/user_server.go
```diff
@@ -85,10 +89,13 @@ func (s *perUIDServer) addConn(a *connectionArgs) error {
 
 func (s *perUIDServer) broadcast(a messageArgs) {
 	var errMsgs []string
+	fmt.Printf("J1\n")
 	for id, conn := range s.conns {
+		fmt.Printf("J2\n")
 		log.Printf("uid %x broadcast to %d", s.uid, id)
 		oc := protocol.OutgoingClient{Cli: rpc.NewClient(conn.xprt, nil)}
 		if err := oc.BroadcastMessage(a.c, a.m); err != nil {
+			// Just log error messages...
```
patrickxb: I think we need to add a log call...I don't see one.

maxtaco: agreed as per comment!

rpc/server.go
```diff
@@ -137,27 +139,33 @@ func (s *Server) confirmUIDShutdown(a confirmUIDShutdownArgs) {
...
 	// matches the last connection in the perUIDServer
 	if serverLast == a.lastConnID {
+		su := s.users[k]
+
 		// remove the perUIDServer from users, lastConns
 		delete(s.users, k)
 		delete(s.lastConns, k)
-		a.ok <- true
+
+		// Non-blocking send that it should self-destruct
```
patrickxb: just to keep comment from other PR here, too:  buffer shutdownCh (1 is ok) instead of the select.

patrickxb: Actually, currently, all the perUser servers have the same shutdownCh, the server's closeCh.  The purpose was to shut down all the perUser servers by closing closeCh, so I think this needs more reworking.

What currently is here could be received by any perUser server (or the main server).

patrickxb: (that's why the type of this channel was `chan struct{}`...to signal that no data was ever meant to be sent to it or received from it, just closing the channel).

maxtaco: Ah ok, can we have a per-UID-thread shutdown channel?

patrickxb: already commited...

patrickxb:
@maxtaco PTAL at the changes.

patrickxb:
Ok, this is definitely an improvement.  I'm having a hard time breaking locally.  See `util/stress.bash` for a script that helped test this.

Ready for another look, @maxtaco

rpc/server.go
```diff
@@ -137,27 +140,31 @@ func (s *Server) confirmUIDShutdown(a confirmUIDShutdownArgs) {
...
+
 		// remove the perUIDServer from users, lastConns
 		delete(s.users, k)
 		delete(s.lastConns, k)
-		a.ok <- true
+
+		// close perUser's selfShutdown channel so it will
+		// self-destruct
+		if su != nil {
+			close(su.selfShutdownCh)
```
maxtaco: oh good thinking! the `close` won't block!

maxtaco:
this looks a lot better!

BUT

I still get a deadlock if I put in a `time.Sleep(10 *time.Millisecond)` above in `getPerUIDServer`

maxtaco:
I spent about an hour looking into it. The problem is that if you insert a `sleep` where I said, the test `TestCloseConnect` fails.  The nature of the failure is that the `CloseListener` channel fires before we've even had a chance to listen for it.  So the notification just drops on the floor and the object is never deleted. I tried various hacks but gave up for the night when my eyes started to bleed!

maxtaco:
See #24 for my thoughts on how to fix the bug found above...

maxtaco:
:+1:

PR #24, opened 2016-04-08
fix deadlock and make deadlocking part of the test

maxtaco:
- the deadlock was due to a race between listening for a close and the close itself.
  if the close came first, the notification was dropped.
- fix that deadlock
- also add more delay slots for testing

### DIFF
--- a/rpc/server_test.go
+++ b/rpc/server_test.go
@@ -44,6 +44,7 @@ func startTestServer(x gregor.NetworkInterfaceIncoming) (*Server, net.Listener, 
 	s := NewServer()
 	s.auth = mockAuth{}
 	s.events = ev
+	s.useDeadlocker = true
 	l := newLocalListener()
 	go s.Serve(x)
 	go s.ListenLoop(l)
--- a/rpc/server_test.go
+++ b/rpc/server_test.go
@@ -286,6 +286,7 @@ func TestCloseConnect(t *testing.T) {
 	}
 
 	<-ev.connCreated
+	<-ev.connDestroyed
 	<-ev.connCreated
 
 	// broadcast a message to goodUID
@@ -295,7 +296,6 @@ func TestCloseConnect(t *testing.T) {
 		t.Logf("broadcast error: %s", err)
 	}
 
-	<-ev.connDestroyed
 	<-ev.bcastSent
 
 	// the user server should still exist due to c2:
--- a/rpc/server_test.go
+++ b/rpc/server_test.go
@@ -286,7 +286,8 @@ func TestCloseConnect(t *testing.T) {
 	}
 
 	<-ev.connCreated
-	<-ev.connDestroyed
+	// taking this out gets rid of the deadlock:
+	// <-ev.connDestroyed
 	<-ev.connCreated
 
 	// broadcast a message to goodUID
--- a/rpc/server_test.go
+++ b/rpc/server_test.go
@@ -286,8 +286,6 @@ func TestCloseConnect(t *testing.T) {
 	}
 
 	<-ev.connCreated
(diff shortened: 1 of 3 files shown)


rpc/user_server.go
```diff
@@ -86,12 +86,23 @@ func (s *perUIDServer) serve() {
...
 	s.conns[a.id] = a.c
 	s.lastConnID = a.id
 	if s.events != nil {
 		s.events.connectionCreated(s.uid)
 	}
+
+	// Well this is cute. We might have missed the s.closeListenCh above
+	// because we started listening too late. So what we might do instead is
+	// check that we're connected, and if not, to send an artificial message
+	// on that channel.
```
patrickxb: We could add a (different?) closeListenCh when the connection is established before the authentication?  It seems like that is the issue:  there's a connection and authentication happens before it goes to the perUID server which listens for close.  But the connection could be closed already.  

Or, at line 89 in this function, check if it is already closed and don't add it.  Perhaps that is simpler.

maxtaco: Over in the RPC repo, @akalin-keybase recommended that we make a `AddCloseListener` before calling Run, with a buffer of 1, and just pass that from place to place.

patrickxb: Nevermind the line 89 idea.

maxtaco: Yeah, I tried that, it didn't work, too much assorted breakage!

patrickxb: Yes, that seems like a better solution (adding the close listener earlier).

rpc/server_test.go
```diff
@@ -285,7 +286,8 @@ func TestCloseConnect(t *testing.T) {
 	}
 
 	<-ev.connCreated
-	<-ev.connDestroyed
+	// taking this out gets rid of the deadlock:
+	// <-ev.connDestroyed
```
patrickxb: @maxtaco this fixes the deadlock, without the change in addConn.

rpc/user_server.go
patrickxb: One thing to note is that the closeListener is shared by all the connections so that the perUIDServer only has to check one channel.  But see my recent commit.  It fixes the test deadlock.

rpc/server_test.go
maxtaco: It doesn't "fix" the deadlock really because the test was right. Now we're just going to be leaking a perUser go-routine in the case that previously caused the deadlock.

maxtaco: Meaning, it's definitely a bug that we're dropping that closedConnection notification in some cases.

patrickxb: The broadcast, however, will notice it.  Or any time another connection is closed.

patrickxb: Moving this event check to after the broadcast works, btw.

maxtaco: But there are still cases in which we never garbage collect the resource.
If the last connection disconnects right away and the user is never used
again.

If this is causing churn it is a pretty corner corner case. Can we leave my
workaround?

patrickxb: Sure.

We might want to implement a resource sweeper to periodically clean things up.  It's possible there are other corner cases we haven't seen yet.

maxtaco: ok, let's make a future issue for it... no need to do it here...

patrickxb:
I re-enabled your workaround.

This LGTM for merging.

maxtaco:
:+1:

maxtaco:
btw, now i'm merged my branch into yours....

patrickxb:
Can we get thumbs up on that one too?

maxtaco:
:+1:

PR #27, opened 2016-04-12
Make connection use rpc.Server.Run()

akalin-keybase:
Also clean up some latent race conditions.


rpc/connection.go
```diff
@@ -61,13 +75,22 @@ func (c *connection) checkAuth(ctx context.Context, m gregor1.Message) error {
 func (c *connection) AuthenticateSessionToken(ctx context.Context, tok gregor1.SessionToken) (gregor1.AuthResult, error) {
 	log.Printf("Authenticate: %+v", tok)
 	res, err := c.parent.auth.AuthenticateSessionToken(ctx, tok)
+	var r authResult
 	if err == nil {
-		c.tok = tok
-		c.uid = res.Uid
-		c.sid = res.Sid
-		c.lastAuthed = c.parent.clock.Now()
+		r = authResult{
```
akalin-keybase: This is racy if somehow we get multiple AuthenticateSessionToken calls

rpc/connection.go
```diff
@@ -23,22 +30,29 @@ type connection struct {
...
+func newConnection(c net.Conn, parent *Server) (*connection, error) {
 	// TODO: logging and error wrapping mechanisms.
 	xprt := rpc.NewTransport(c, nil, nil)
 
 	conn := &connection{
 		c:      c,
 		xprt:   xprt,
 		parent: parent,
-		authCh: make(chan error, 100),
+		authCh: make(chan authResult, 1),
```
akalin-keybase: there's no reason for this to store more than 1, right? We can just ignore subsequent ones, anyway (See below).

akalin-keybase:
Not sure what the test coverage is like for this code, but I can add some tests if needed.

rpc/connection.go
```diff
@@ -106,28 +129,44 @@ func (c *connection) startRPCServer() <-chan error {
...
+	case err := <-c.closeCh:
+		// Always non-nil.
 		return err
+
+	case r := <-c.authCh:
+		if r.err != nil {
+			return r.err
+		}
+
+		if r.res.Uid == nil {
```
jacobhaven: This should never happen.

rpc/connection.go
```diff
@@ -106,28 +129,44 @@ func (c *connection) startRPCServer() <-chan error {
...
 	}
-	if c.uid == nil {
-		return ErrBadUID
-	}
-	return nil
 }
 
 func (c *connection) close() {
-	close(c.authCh)
+	// Should trigger the c.closeCh case in startAuthentication.
```
jacobhaven: It doesn't? You need to `close(c.closeCh)`. That will set `err=nil`, though.

rpc/connection.go
```diff
@@ -61,13 +75,22 @@ func (c *connection) checkAuth(ctx context.Context, m gregor1.Message) error {
...
+			time: c.parent.clock.Now(),
+		}
+	} else {
+		r = authResult{err: err}
+	}
+	select {
+	case c.authCh <- r:
+		// First auth call.
+	default:
+		// Subsequent auth calls -- just drop.
```
jacobhaven: This isn't valid. We need to handle all auth calls, as later ones might fail due to timeout or revocation.

rpc/connection.go
akalin-keybase: calling Close() will cause srv.Run() to return, which will be sent onto closeCh

rpc/connection.go
```diff
@@ -106,28 +129,44 @@ func (c *connection) startRPCServer() <-chan error {
...
-			return errCh
+			return err
 		}
 	}
 
-	return srv.RunAsync()
+	closeCh := make(chan error, 1)
+	go func() {
+		err := srv.Run()
+		closeCh <- err
```
jacobhaven: Why not `closeCh <- srv.Run()`?

rpc/connection.go
akalin-keybase: Okay, removed. (The old code had this, too, FWIW.)

rpc/connection.go
akalin-keybase: Fixed.

rpc/connection.go
akalin-keybase: Now moot (with upstream RPC changes)

akalin-keybase:
CI will fail until https://github.com/keybase/go-framed-msgpack-rpc/pull/75 is merged (which I don't want to do until this is LGTMed, to avoid breakage...)

jacobhaven:
Some of these changes are dependent on https://github.com/keybase/go-framed-msgpack-rpc/pull/75 merging first (thus the CI failure), but otherwise looks good.

jacobhaven:
This LGTM once https://github.com/keybase/go-framed-msgpack-rpc/pull/75 is merged.

PR #28, opened 2016-04-13
consistent error wrap and unwrap functions in all places

maxtaco:
- requires maxtaco/CORE-2878 in keybase/client
- sadly pulls in a keybase1 dependency but that's better than libkb


maxtaco:
Requires https://github.com/keybase/client/pull/2599

patrickxb:
LGTM

PR #29, opened 2016-04-13
fix the AuthenticateSessionToken rpc

oconnor663:
The name of the parameter was out-of-sync with what we have in kbweb.

r? @maxtaco


oconnor663:
Tests pass on my machine.

maxtaco:
:+1: when CI says green

PR #34, opened 2016-04-14
Add InBandMessagesSince to RPC and NetworkInterfaceIncoming

jacobhaven:
r? @maxtaco @patrickxb


protocol/gregor1/avdl/incoming.avdl
```diff
@@ -3,4 +3,5 @@
 
 protocol incoming {
 	void consumeMessage(Message m);
+	array<InBandMessage> inBandMessagesSince(UID u, DeviceID d, Time t);
```
patrickxb: u, d, and t seem a little terse for protocol parameters...

jacobhaven: Fair, *updated*

jacobhaven:
It's a bit annoying that we have `gregor1.IncomingInterface` and `gregor.NetworkInterfaceIncoming`, which are basically the same, except for the protocol/interface type of the args. Should we consolidate onto just one? (I don't particularly care which)

maxtaco:
Happy with consolidating.  We have the same issue in some places in the main client too

protocol/gregor1/avdl/incoming.avdl
```diff
@@ -3,4 +3,5 @@
 
 protocol incoming {
 	void consumeMessage(Message m);
+	array<InBandMessage> inBandMessagesSince(UID uid, DeviceID deviceid, Time ctime);
```
maxtaco: `TimeOrOffset` is better for clients that don't trust their clock

jacobhaven: They should request from the last known `CTime` (that was set for them by the server). If they don't have that, they should request everything (send `ctime=0`)

jacobhaven:
I just uncoupled a separate `sync.avdl`/`SyncInterface` to avoid touching `gregor.NetworkInterfaceIncoming`

storage/sql_sm.go
```diff
@@ -483,6 +502,26 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t grego
...
+	}
+	arg := gregor1.InBandMessagesSinceArg{Ctime: gregor1.ToTime(ctime)}
+	if u != nil {
+		arg.Uid = gregor1.UID(u.Bytes())
+	}
+	if d != nil {
+		arg.Deviceid = gregor1.DeviceID(d.Bytes())
+	}
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	newMsgs, err := s.sync.InBandMessagesSince(ctx, arg)
```
maxtaco: why this?

jacobhaven: ? It's not recursive. It calls out an RPC to get the messages not stored in the `StateMachine` (appended to the ones from the `StateMachine`)

maxtaco:
I think this is going in the right direction. Can you add some tests? Also can you assert after the catch up that you have the same state that the server does?

jacobhaven:
I've refactored a lot of things.
* Renamed the RPC to just `Sync` instead of `InBandMessagesSince` to avoid confusion with the `StateMachine` method.
* Removed Networking Interfaces from "gregor", so they could be handled entirely by the rpc protocol "gregor1"
* Added the server-side plumbing for `Sync` through `connection` and `Server` 
* Fleshed out client tests, checking hashes of State against the server and restored client.

storage/sql_sm.go
```diff
@@ -462,7 +462,7 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t time.
 		qb.Build("AND (m.devid=? OR m.devid IS NULL)", hexEnc(d))
 	}
 
-	qb.Build("AND m.ctime >= ")
+	qb.Build("AND m.ctime > ")
```
jacobhaven: I'm pretty sure these are the semantics we want. Otherwise we end up duplicating the last message (a similar change was made in `MemEngine`)

maxtaco: I think it was right as is, since you could get two messages in the same microsecond, but you only saw one and then you hung up.  The protocol should be idempotent / resilient to repeated messages.

In fact, I was thinking, just to make sure, we'd ask for messages since (last-ctime - 1 minute) or so, since messages can be reordered, and we might have missed some late messages as we were shutting down.

storage/client.go
```diff
@@ -62,3 +68,40 @@ func (c *Client) Restore() error {
...
+	var ctime time.Time
+	if t := c.sm.LatestCTime(); t != nil {
+		ctime = *t
+	}
+
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	arg := gregor1.SyncArg{
+		Uid:      gregor1.UID(c.user.Bytes()),
+		Deviceid: gregor1.DeviceID(c.device.Bytes()),
+		Ctime:    gregor1.ToTime(ctime),
```
maxtaco: what to you think about `ctime - time.Minute` (as per earlier comment) ?

storage/sql_sm.go
```diff
@@ -453,8 +462,8 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t grego
 		qb.Build("AND (m.devid=? OR m.devid IS NULL)", hexEnc(d))
 	}
 
-	qb.Build("AND m.ctime >= ")
-	qb.TimeOrOffset(t)
+	qb.Build("AND m.ctime > ")
```
maxtaco: see earlier comment

storage/client_test.go
```diff
@@ -37,4 +38,8 @@ func TestLevelDBClient(t *testing.T) {
 	if err := c.Restore(); err != nil {
 		t.Fatal(err)
 	}
+
+	if err := c.Sync(); err != nil {
```
maxtaco: does this actually get any new messages?  don't you need to generate some new ones to assert that you're using the `Sync` RPC?

storage/mem_sm.go
```diff
@@ -311,12 +312,24 @@ func (m *MemEngine) State(u gregor.UID, d gregor.DeviceID, t gregor.TimeOrOffset
...
 
-func (m *MemEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t gregor.TimeOrOffset) ([]gregor.InBandMessage, error) {
+func (m *MemEngine) LatestCTime() (latestCTime *time.Time) {
 	m.Lock()
 	defer m.Unlock()
-	user := m.getUser(u)
-	msg := user.replayLog(m.clock.Now(), d, t)
-	return msg, nil
+	for _, u := range m.users {
+		_, ctime := u.replayLog(m.clock.Now(), nil, time.Time{})
```
maxtaco: Will this be expensive on mobile devices on startup, to iterate through all existing messages to find out when the last one was?

storage/sql_sm.go
```diff
@@ -378,7 +378,7 @@ func (s *SQLEngine) inBandMetadataSince(u gregor.UID, t gregor.TimeOrOffset) ([]
 	qb := s.newQueryBuilder()
 	qb.Build(qry, hexEnc(u))
 	if t != nil {
-		qb.Build("AND ctime >= ")
+		qb.Build("AND ctime > ")
```
maxtaco: Also still think `>=` is better here....

storage/sql_sm.go
```diff
@@ -435,7 +435,16 @@ func (s *SQLEngine) rowToInBandMessage(u gregor.UID, rows *sql.Rows) (gregor.InB
 	return nil, nil
 }
 
-func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t gregor.TimeOrOffset) ([]gregor.InBandMessage, error) {
+func (s *SQLEngine) LatestCTime() *time.Time {
+	var ctime timeScanner
+	err := s.driver.QueryRow(`SELECT ctime FROM messages ORDER BY m.ctime DESC`).Scan(&ctime)
```
maxtaco: Can you add `LIMIT 1`?

maxtaco: Actually, can you do `SELECT MAX(ctime) FROM message`?

test/mocks.go
```diff
@@ -0,0 +1,44 @@
...
+	Consumed int
+	SM       gregor.StateMachine
+}
+
+func (m *MockIncomingInterface) ConsumeMessage(_ context.Context, msg gregor1.Message) error {
+	m.Consumed++
+	return m.SM.ConsumeMessage(msg)
+}
+
+func (m *MockIncomingInterface) Sync(_ context.Context, arg gregor1.SyncArg) (res gregor1.SyncResult, err error) {
```
maxtaco: this mock seems very similar to the actual code.  is there any way to consolidate the two?

storage/sql_sm.go
jacobhaven: Seems reasonable, *done*.

storage/mem_sm.go
jacobhaven: I can't really think of a better way to handle this without maintaining the `user` log as a list ordered by `ctime`. Maybe that is the way to go...

storage/client_test.go
jacobhaven: The right thing to do here is to create two separate `StateMachine`s, give them different states and `Sync` between them.

test/mocks.go
jacobhaven: *done*

storage/sql_sm.go
jacobhaven: *reverted*

storage/client.go
jacobhaven: I don't know what earlier comment you're referring to. Now that I've added `MsgID` deduplication code to `MemEngine`, this should work fine.

storage/client.go
```diff
@@ -62,3 +68,40 @@ func (c *Client) Restore() error {
...
+		return err
+	}
+
+	hash, err := state.Hash()
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Equal(res.Hash, hash) {
+		return errors.New("local state hash != server state hash")
```
jacobhaven: In this case, we should discard our local state and replace it with a `Sync` from time `0` (if the hash fails after that, something is really wrong and we can error out.)

storage/mem_sm.go
jacobhaven: Actually, we don't need the *precise* `LatestCTime`, so we can probably get away with iterating backwards until we find the last message with the correct user/device IDs.

storage/sql_sm.go
```diff
@@ -435,9 +435,12 @@ func (s *SQLEngine) rowToInBandMessage(u gregor.UID, rows *sql.Rows) (gregor.InB
 	return nil, nil
 }
 
-func (s *SQLEngine) LatestCTime() *time.Time {
+func (s *SQLEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) *time.Time {
+	qry := `SELECT MAX(ctime) 
+			FROM messages
+			WHERE uid = ? AND devid = ?`
```
maxtaco: `(devid = ? OR devid IS NULL)` since a `NULL` devid means "for all devices"

storage/mem_sm.go
```diff
@@ -318,11 +318,18 @@ func (m *MemEngine) State(u gregor.UID, d gregor.DeviceID, t gregor.TimeOrOffset
...
-func (m *MemEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) (latestCTime *time.Time) {
+func (m *MemEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) *time.Time {
 	m.Lock()
 	defer m.Unlock()
-	_, latestCTime = m.getUser(u).replayLog(m.clock.Now(), d, time.Time{})
-	return
+	log := m.getUser(u).log
+	for i := len(log) - 1; i >= 0; i-- {
+		if log[i].i != nil && log[i].i.item != nil &&
+			(d == nil || (log[i].i.item.Metadata() != nil && log[i].i.item.Metadata().DeviceID() != nil &&
```
maxtaco: if `log[i].i.item.Metadata().DeviceID() == nil` then this message counts too, since `nil` deviceID means "all devices"

storage/client.go
jacobhaven: **Done**

storage/mem_sm.go
jacobhaven: **updated**

maxtaco: Thanks! Lemme just take a quick read through the test before I give it a
thumb. I hope to get to it a little later tonight.

storage/sql_sm.go
```diff
@@ -457,7 +457,7 @@ func (s *SQLEngine) Clear() error {
 func (s *SQLEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) *time.Time {
 	qry := `SELECT MAX(ctime) 
 			FROM messages
-			WHERE uid = ? AND devid = ?`
+			WHERE uid = ? AND devid = ? OR devid IS NULL`
```
maxtaco: I'm pretty sure you need parens here!  `AND` binds tighter than `OR`.

storage/client_test.go
maxtaco: Does this check that the state winds up being the same across the two devices (maybe via hash?)

storage/client.go
```diff
@@ -62,3 +69,55 @@ func (c *Client) Restore() error {
...
+		return errHashMismatch{}
+	}
+
+	return nil
+}
+func (c *Client) Sync() error {
+	if err := c.syncFromTime(c.sm.LatestCTime(c.user, c.device)); err != nil {
+		if _, ok := err.(errHashMismatch); ok {
+			log.Printf("Sync failure: %v\nReseting StateMachine and retrying")
+			//c.sm.Clear()
```
maxtaco: Delete this line?

jacobhaven: Oh, right. *uncommented*

storage/sql_sm.go
jacobhaven: **fixed**

storage/client_test.go
```diff
@@ -106,6 +106,17 @@ func TestLevelDBClient(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	if err := c.syncFromTime(c.sm.LatestCTime(c.user, c.device)); err != nil {
+		t.Fatal(err)
+	}
+
```
maxtaco: Can you explain why calling `AddStateMAchinePerDevice` causes `syncFromTime` to fail?

maxtaco:
👍

PR #35, opened 2016-04-14
Fix missing return in tryShutdown

patrickxb:
r? @maxtaco 

The fix is in tryShutdown(), it was missing a return if connections were added since the shutdown request was added.

Tested with 3ns, 3ms, test -count 10000, stress.bash, stress.bash with race.

### DIFF
--- a/rpc/server_test.go
+++ b/rpc/server_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sync"
 	"testing"
 
 	keybase1 "github.com/keybase/client/go/protocol"
@@ -106,6 +107,7 @@ type client struct {
 	cli        *rpc.Client
 	broadcasts []gregor1.Message
 	shutdown   bool
+	sync.Mutex
 }
 
 func newClient(addr net.Addr) *client {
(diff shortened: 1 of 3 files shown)


rpc/user_server.go
```diff
@@ -128,12 +128,13 @@ func (s *perUIDServer) broadcast(a messageArgs) {
...
 
-// tryShutdown checks if it is ok to shutdown.  Returns true if it
-// is ok.
+// tryShutdown makes a request to the parent server for this
+// server to be terminated.
 func (s *perUIDServer) tryShutdown() {
 	// make sure no connections have been added
 	if len(s.conns) != 0 {
 		log.Printf("tried shutdown, but %d conns for %x", len(s.conns), s.uid)
+		return
```
patrickxb: This is the fix.

rpc/server_test.go
```diff
@@ -106,6 +107,7 @@ type client struct {
 	cli        *rpc.Client
 	broadcasts []gregor1.Message
 	shutdown   bool
+	sync.Mutex
```
patrickxb: This fixes a race found by race detector.

rpc/server_test.go
```diff
@@ -133,6 +135,8 @@ func (c *client) Shutdown() {
 	c.conn.Close()
 	// this is required as closing the connection only closes one direction
 	// and there is a race in figuring out that the whole connection is closed.
+	c.Lock()
+	defer c.Unlock()
```
maxtaco: man, sometimes I just don't know about Go....

maxtaco:
Great find!! :+1: :+1:

PR #36, opened 2016-04-14
create the ping interface in incoming.avdl

oconnor663:
This will be the first thing we hit in integration tests.

r? @patrickxb @maxtaco

### DIFF
--- a/protocol/gregor1/avdl/incoming.avdl
+++ b/protocol/gregor1/avdl/incoming.avdl
@@ -3,4 +3,5 @@
 
 protocol incoming {
 	void consumeMessage(Message m);
+	string ping();
 }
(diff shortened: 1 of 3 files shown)


patrickxb:
LGTM

protocol/gregor1/avdl/incoming.avdl
```diff
@@ -3,4 +3,5 @@
 
 protocol incoming {
 	void consumeMessage(Message m);
-}
\ No newline at end of file
+	string ping();
```
jacobhaven: Why not return `void`?

oconnor663: Copying from Slack: No huge reason. I guess it makes it slightly less likely that we overlook a failure, if `Ping()` returns nil for some other reason.

PR #37, opened 2016-04-15
create a Dockerfile, and make Travis push images

oconnor663:
creating PR to trigger Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,18 +11,31 @@ addons:
     - mysql-server-5.6
     - mysql-client-core-5.6
     - mysql-client-5.6
-before_install:
-  - mysql -e "create database IF NOT EXISTS test;" -uroot
 env:
   - TEST_MYSQL_DSN=root:@/test 
 notifications:
   slack:
     rooms:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,18 +11,31 @@ addons:
...
-  - (cd protocol && npm i && git diff --quiet --exit-code)
 matrix:
   - os: osx
 branches:
   only:
     - master
+services:
+  - docker
+before_install:
+  - mysql -e "create database IF NOT EXISTS test;" -uroot
```
jzila: is this a local mysql install?

.travis.yml
```diff
@@ -11,18 +11,31 @@ addons:
...
+before_script:
+  - (cd protocol && npm i && git diff --quiet --exit-code)
```
jzila: what does this do?

.travis.yml
oconnor663: I think so. I just moved this line from above, to make the order of operations look clearer to me.

.travis.yml
oconnor663: This is checking to make sure that the generated-and-checked-in protocol files haven't gotten out of date compared to the AVDL files they're generated from. This is another section that I just moved down from above.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:14.04
+MAINTAINER Keybase <admin@keybase.io>
+
+RUN useradd -m keybase
+USER keybase
+ENV PATH=/home/keybase:$PATH
+WORKDIR /home/keybase
+
+COPY ./gregord/gregord /home/keybase/gregord
```
jzila: This assumes `go build` and not `go install`, correct?

jzila:
👍

Dockerfile
oconnor663: Yep, that's the `(cd gregord && go build)` above.

PR #38, opened 2016-04-18
Update MySQL DSN parsing from S3 bucket

jacobhaven:


jacobhaven:
Similar to [code here](https://github.com/keybase/kbfs/blob/2760c2a2e3201a194b4487dae40c1f1f2af1eed4/serverutil/config.go#L309-L324)

maxtaco:
👍

PR #43, opened 2016-04-21
Sync Bugfix (2)

jacobhaven:
This was the missing piece in why `Sync` was crashing. r? @maxtaco @patrickxb @oconnor663 
(thanks @aalness for giving me access to the docker logs which had a very stack track)


patrickxb:
LGTM

PR #44, opened 2016-04-21
Vendored dependencies only for gregord

jzila:
Since `gregord` is an app that is built, I vendored its dependencies only, while leaving the rest of the packages unvendored. That way if someone wants to build `gregord`, there's no dependency ambiguities, but they can include the libraries against their own vendoring.

review please @jacobhaven, cc @maxtaco


jacobhaven:
I love vendoring PRs, you can sneak in all the malicious code you want :)

👍

jzila:
@jacobhaven sort of...the HEAD commit IDs are stored in vendor.json, and those get verified. Maybe we could put in a check via [govendor](https://github.com/kardianos/govendor/blob/fetch/doc/design.md) to ensure that the tree matches the official package at the commit specified in `vendor.json`.

PR #45, opened 2016-04-21
Fix broken TestLevelDBClient

jacobhaven:
r? @jzila @maxtaco @patrickxb 

Caused by #43, which passed CI before being merged... odd.


jzila:
🚢

PR #48, opened 2016-04-21
set up logrus, with optional env vars for syslog

oconnor663:
r? @jacobhaven 

Also @jzila to sanity check the vendoring.

Note that gregord vendors its own copy of the rpc directory, so we have
to make these changes twice. This in turn is because vendoring things
properly at the top level of the gregor project causes conflicts, when
other projects that vendor gregor have different versions of the same
dependencies. This is a nightmare.


oconnor663:
I got very frustrated with bugs and limitations in both go-logging and logrus, and I ended up almost flipping a coin. I'm happy to bikeshed this some more if anyone wants to :|

oconnor663:
(One of the downsides of go-logging is that you need to declare `var logger = ...` or something similar in each package. In `main` it's pretty clear that this belongs in `main.go`, but in other packages there's no obvious place to put it, and I end up feeling very dirty about the whole thing :p Very arbitrary on my part, I admit.)

jzila:
the vendoring seems fine to me.

maxtaco:
Jack how will this integrate with the main client when Gregor is run as a
library?

jzila:
@maxtaco you mean with respect to vendoring? `gregord` is a binary that is separate from the library. Since the library is unvendored, it should continue to function correctly when imported elsewhere.

maxtaco:
No I mean that we need to call into Gregor library code from inside the
service process.  Where will the logs wind up?

maxtaco:
(Assuming that we can compile it at all despite the vendoring issues.  Btw
we had previously vendored in Gregor and then ripped it out because we
couldn't compile the Gregor libraries from the service. But I assume you
found some answer for this).

The high bit is that both Gregord and keybase service need acces to the
Gregor libraries. So logging has to work in both settings.

gregord/vendor/github.com/keybase/gregor/rpc/connection.go
```diff
@@ -70,8 +70,8 @@ type connection struct {
 }
 
 func newConnection(c net.Conn, parent *Server) (*connection, error) {
-	// TODO: logging and error wrapping mechanisms.
-	xprt := rpc.NewTransport(c, nil, keybase1.WrapError)
+	// TODO: error wrapping mechanisms.
+	xprt := rpc.NewTransport(c, NewLogrusRPCLogFactory(), keybase1.WrapError)
```
maxtaco: Error wrapping is already done

gregord/vendor/github.com/keybase/gregor/rpc/connection.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"log"
+	log "github.com/Sirupsen/logrus"
```
maxtaco: Will we still get log rotation in keybase client and kbfs client if we are outputting rpc trace via log?

jzila:
@maxtaco yeah, originally we vendored gregor at the root. Now we only vendor `gregord`, which shouldn't ever be used as a library anywhere.

oconnor663:
@maxtaco I was imagining that the client `main` function would configure logrus (or go-logging, whatever) globally on startup, with a writer doing whatever it needs to do, possibly kicking off a goroutine behind that writer to forward everything to `G.Log`. Now that you mention it, though, does the client need this to be done on a per-connection basis rather than globally? Does that require that we pass logger objects of some kind all the way down through every library we call?

gregord/vendor/github.com/keybase/gregor/rpc/connection.go
```diff
@@ -214,3 +214,29 @@ func (c *connection) close() {
 var _ gregor1.AuthInterface = (*connection)(nil)
 
 // var _ gregor1.IncomingInterface = (*connection)(nil)
```
jacobhaven: Hmm, this should probably be uncommented.

gregord/vendor/vendor.json
```diff
@@ -66,8 +66,8 @@
 		},
 		{
 			"path": "github.com/keybase/gregor/rpc",
-			"revision": "981e40adf9373075abd789255a21b2e56ec57ffb",
-			"revisionTime": "2016-04-21T10:43:38-07:00"
+			"revision": "030769e7b0d669525d277a3b4e9c3a327fa84721",
```
jacobhaven: You should also add `github.com/Sirupsen/logrus`

maxtaco:
Ok great all set there.

Now the question is where do the logs go when linked as a library.

maxtaco:
I guess we can do all of that but then what is logrus actually providing then? and I would be curious to know if such plumbing can actually work.  Does it pope through properly and pick up log rotation, and work in C/I, etc....

maxtaco:
I have no idea what "pope" was supposed to mean above or how to edit a previous comment from my phone.

maxtaco:
Maybe you need to provide a logrus backend that writes to go-logging :(

oconnor663:
@maxtaco I would prefer not to have anything in the gregor process be responsible for log rotation. In fact, nothing in the gregor process is currently aware of any log file paths at all. Right now it just prints to stdout, and optionally also logs to syslog (which means sending UDP packets to a host we specify through env vars). Whatever's on the receiving end of our syslog packets will take care of rotation, or possibly not store logs on disk much at all.

I was reluctant to even add the syslog details in-process, but it buys us a few things (I could've saved myself a lot of trouble by just listening to @jacobhaven on this one...but now I understand [the syslog protocol](https://tools.ietf.org/html/rfc5424)):

- a single log entry can span multiple lines without getting chopped up into multiple logical entries
- a standardized (not _great_, but standardized :p) way to encode info/warn/error levels
- a workaround for the fact that Docker's stdpout/stderr logs don't rotate very well right now (though I haven't yet actually killed our regular printing when syslog is enabled)

The downsides of not relying exclusively on stdout/stderr are:

- we need some other story for logging stacktraces when our service panics
- random 3rd party dependencies that decide to print to stdout/stderr will get missed

This is all in contrast to what we do on the client, where our process writes logs directly to files that it knows about. That does help us make `keybase log send` work in a more platform-neutral way, but it's also causing us a lot of problems when we end up in weird daemon autospawning situations.

patrickxb:
I don't know docker, but I'm a big fan of just logging to stdout and letting whatever runs the server process handle the log messages however it wants to.  

It's very flexible and allows logging backends to be changed without code changes.

jzila:
I will add this: @aalness built the mdserver to log json lines, which gives us all the advantages of syslog. That said, I'm not at all an expert on this stuff.

jacobhaven:
If we've decided we're using logrus and `NewLogrusRPCLogFactory()`, can we go ahead and merge this to base other code on? Then we can followup with more adding a bunch of specific logging throughout gregor and deciding how we want to handle log rotation/etc. when using gregor as a library?

oconnor663:
Max repeated his main concern to me over Slack, which is that we have to have a good story for how all of this integrates into the client's logging when the client calls into libgregor code. The client daemon's logging is _not_ process global, but rather contextified per request. That has a kind of viral property, where everything we call into *also* needs to be able to contextify its logging, or else the client will permanently have to resort to nasty hacks to slurp global log lines back into whatever (more than one) contexts happen to be alive at that time.

That makes this PR something of a big decision: either we will do what we have here and doom the client to hackiness, or we will have to pass _some_ kind of object as an explicit arg to _every_ struct/function we define in libgregor (and in future libraries presumably), as we do in libkb. I'm leaning towards the hackiness, but I want to be clear about that.

oconnor663:
@jacobhaven's instinct is to go with implicit passed-from-above log objects. I'm up for that, but it'll probably mean adding a dependency on `github.com/keybase/client/go/logger`. @jzila are you ok with that? :)

jzila:
Is there a way we could break out `client/go/logger` into its own library?

aalness:
re: syslog/stdout. Seems mostly orthogonal to application logging libraries and interfaces. We can play with the Docker log drivers or other mechanisms later to achieve whatever we think is optimal.

jzila:
It sounds to me like Docker here is a motivating factor, but overall the discussion is whether we want to log contextually to a service designed for it, or whether we want to continue with text-based logging that we then have to parse somehow.

oconnor663:
Agreed with @aalness. The two questions of "global vs contextified" and "stdout vs syslog" are independent.

@jzila, did you mean it's own repo? I think at least for the purposes of this PR, I want to keep it where it is, and then vendor it into `gregor/gregord/` as we've done with other stuff. As far as Go is concerned, I think the subdirectory makes it its own library already?

jzila:
@oconnor663 I mean abstract out `client/go/logger`, and import that in both gregor and client.

oconnor663:
It's currently abstracted. Strib did a lot of that a while back, and we have a Logger interface that would be the only thing we'd depend on in most calling code.

I'm going to put up a separate PR with that approach, which we'll probably go with instead of this one. More soon.

oconnor663:
https://github.com/keybase/gregor/pull/50 is up. It will probably replace this.

PR #49, opened 2016-04-22
Lazy Notification protocol

jacobhaven:
r? @maxtaco


storage/mem_sm.go
```diff
@@ -347,6 +357,70 @@ func (m *MemEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t time.
...
+					}
+					heap.Push(&nq, not)
+				}
+			}
+		}
+	}
+
+	return nq, nil
+}
+
```
jacobhaven: It might be best to consilidate these implemention int he `MemEngine` and `SQLEngine`

```Go
func (nq notificationQueue) sendNotifications(addNotificationCh <-chan gregor.Notification, notificationCh chan <-gregor.Notification, cl clockwork.Clock, cleanup func())
```

notifyd/config.go
```diff
@@ -0,0 +1,134 @@
...
+
+func warnf(f string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, f, args...)
+}
+
+func errorf(f string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, "error: "+f, args...)
+}
+
+func parseMysqlDSN(raw *rawOpts) (*url.URL, error) {
```
maxtaco: I recognize this code.  If it's copied from somewhere else, can we factor it out into a library?

notifyd/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+func main() {
+	opts, err := ParseOptions(os.Args)
+	if err != nil {
+		log.Fatal(err)
+	}
+	Cli := rpc.NewClient(rpc.NewTransport(conn, nil, keybase1.WrapError), keybase1.ErrorUnwrapper{})
+	n := newNotifier(opts.MysqlDSN, gregor1.NotifyClient{Cli})
+	defer n.shutdown()
+
```
maxtaco: do we need to handle signals explicitly here?

protocol/gregor1/avdl/notify.avdl
```diff
@@ -0,0 +1,4 @@
+@namespace("gregor.1")
+protocol notify {
+   void notify(ItemAndMetadata md);
```
maxtaco: the API server will answer this, right?

storage/sql_sm.go
```diff
@@ -126,7 +129,7 @@ func (s *SQLEngine) consumeCreation(tx *sql.Tx, u gregor.UID, i gregor.Item) err
 			continue
 		}
 		nqb := s.newQueryBuilder()
-		nqb.Build("INSERT INTO items(uid, msgid, ntime) VALUES(?,?,", hexEnc(u), hexEnc(md.MsgID()))
+		nqb.Build("INSERT INTO reminders(uid, msgid, ntime) VALUES(?,?,", hexEnc(u), hexEnc(md.MsgID()))
```
maxtaco: whooops!

protocol/gregor1/avdl/notify.avdl
jacobhaven: Yep.

storage/sql_sm.go
```diff
@@ -517,6 +539,68 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t time.
 	return ret, nil
 }
 
+func (s *SQLEngine) prepopulateNotifications() (notificationQueue, error) {
+	var nq notificationQueue
+	qry := `SELECT i.msgid, m.devid, i.category, i.dtime, i.body, m.ctime
```
maxtaco: Don't you need to limit to items whose notification time has come?

notifyd/config.go
jacobhaven: Already done: https://golang.org/pkg/flag/

maxtaco:
Seems like a great start.  Biggest question surrounding selecting the items to be notified from the DB.

notifyd/config.go
maxtaco: Is that link correct?

notifyd/main.go
jacobhaven: I'm not sure what you mean. If we receive a `SIGINT`/`SIGTERM`/etc., all the cleanup in `shutdown` will happen automatically. The `shutdown` is really just here for style, to match what should be used in the tests.

maxtaco: Cool, I didn't know that about go... Good to know

notifyd/config.go
maxtaco: Oh cool, good idea..

storage/sql_sm.go
jacobhaven: No, the point is to initialize a sorted `notificationQueue` of all future notifications already known in the DB, and then push into it whenever a new reminder is added. I don't think the memory footprint should be too large here.
Otherwise we'd be constantly querying the DB to determine what to send on the channel, and they'd be out-of-date.

maxtaco: I worry this approach won't scale.  You might be fetching millions of rows here.

jacobhaven:
I think we should change this to more of a batch approach. Have one `notifyd` (maybe multiple if we do DB sharding) that periodically (once a minute, once an hour?) queries the DB for all `reminders` whose time is past (associated with a non-dismissed item) and sends a batch of notifies. The protocol can even change to `void notify(array<ItemAndMetadata> md);`

maxtaco:
Sgtm

storage/sql_sm.go
```diff
@@ -517,6 +536,32 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t time.
...
+	}
+
+	var reminders []gregor.Reminder
+	for rows.Next() {
+		not, err := s.rowToReminder(rows)
+		if err != nil {
+			return nil, err
+		}
+		reminders = append(reminders, not)
+	}
```
jacobhaven: Should delete entries from `reminders` table before returning.

reminderd/config.go
```diff
@@ -0,0 +1,184 @@
+package main
```
jacobhaven: We should handle Option parsing for `reminderd` and `gregord` in a more general way.

reminderd/main.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	conn, err := opts.RemindServer.Dial()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	Cli := rpc.NewClient(rpc.NewTransport(conn, nil, keybase1.WrapError), keybase1.ErrorUnwrapper{})
```
maxtaco: use the same logging as in the new logging PR....

reminderd/reminder.go
```diff
@@ -0,0 +1,61 @@
...
+	"golang.org/x/net/context"
+)
+
+type reminder struct {
+	db     *sql.DB
+	sm     gregor.StateMachine
+	remind gregor1.RemindInterface
+}
+
+func newReminder(dsn *url.URL, remind gregor1.RemindInterface) (*reminder, error) {
```
maxtaco: "reminder" can mean both one who reminds, and the thing the one who reminds does.  Can you use a different word for one of the two?

storage/sql_sm.go
```diff
@@ -517,6 +536,50 @@ func (s *SQLEngine) InBandMessagesSince(u gregor.UID, d gregor.DeviceID, t time.
...
+
+	del := `DELETE FROM reminders WHERE ntime <=`
```
maxtaco: I think it's safe to remove them one-by-one after they have been successfully sent.  In this scheme, we can delete a whole batch and send none of them, since the receiver is down above...

reminderd/reminder.go
```diff
@@ -0,0 +1,61 @@
...
+
+	var rs1 []gregor1.Reminder
+	for _, rm := range rs {
+		if rm, ok := rm.(gregor1.Reminder); ok {
+			rs1 = append(rs1, rm)
+		}
+	}
+
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	return r.remind.Remind(ctx, rs1)
```
maxtaco: in the current plan, the reminders are deleted from the DB regardless of whether this RPC works... should only remove after the reminder has been ACKed.  see comment below.

reminderd/reminder.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	var rs1 []gregor1.Reminder
+	for _, rm := range rs {
+		if rm, ok := rm.(gregor1.Reminder); ok {
+			rs1 = append(rs1, rm)
+		}
+	}
+
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
```
maxtaco: why such a short timeout?

interface.go
```diff
@@ -88,6 +88,11 @@ type Item interface {
 	Category() Category
 }
 
+type Reminder interface {
+	Item() Item
+	NotifyTime() time.Time
```
maxtaco: can you `s/Notify/Remind` everywhere? I agree it's a better word, since `Notify` is used to mean something else elsewhere.

reminderd/config.go
```diff
@@ -0,0 +1,184 @@
...
+func warnf(f string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, f, args...)
+}
+
+func errorf(f string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, "error: "+f, args...)
+}
+
+func parseMysqlDSN(raw *rawOpts) (*url.URL, error) {
+	var dsn string
```
maxtaco: i asked previously, but can this be factored out, since the same code is elsewhere in the project.

reminderd/config.go
```diff
@@ -0,0 +1,184 @@
...
+		return badUsage("No mysql-dsn specified")
+	}
+
+	return nil
+}
+
+// readFromS3Config reads the content of the files denoted by fileNames
+// from S3 configuration bucket
+func readFromS3Config(region string, bucket string, fileNames ...string) ([]string, error) {
+	if _, ok := aws.Regions[region]; !ok {
```
maxtaco: this code is also elsewhere in the project...

reminderd/reminder.go
jacobhaven: I couldn't think of one.

reminderd/main.go
jacobhaven: It just got merged, I haven't rebased off of it yet.

interface.go
jacobhaven: I did for most of them, but kept `NotifyTime()` to match `ntime`.

reminderd/config.go
maxtaco: oh there's a comment above.  agree.

reminderd/reminder.go
maxtaco: a `nagger` is one who reminds a lot...  @chrisnojima suggests `alarm` and `alarmer` . @cjb seconds.

jacobhaven:
@maxtaco could you take another look? I've restructured things so that gregord, notifyd + any future daemons can share code and vendoring. I've also delayed reminder deletion from the DB until a successful `remind` RPC call is sent.

reminderd/reminder.go
chrisnojima: Maybe "alarmist" vs alarmer

daemons/naggerd/nagger.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+	if err := n.remind.Remind(ctx, reminders1); err != nil {
+		return err
+	}
+
+	for _, rm := range reminders {
+		if err := n.sm.DeleteReminder(rm); err != nil {
```
maxtaco: We now have the opposite concern, that the nag is sent but not deleted from the DB. We'll have to make sure we're not crashing too much, because the worst case scenario would be some sort of flapping in which the nagger comes up, issues a bunch of reminders, and then NPEs (or whatnot) on the first delete, and then comes back up and nags the same group of users again.

I'm not sure of what the right answer is here.  I know I asked for this change, but I'm not quite so sure now.  We can have a last-minute per-email email bomb prevention mechanism over on the mail spooler side.

Ideas?

maxtaco:
Looking really good. How about testing?  Also, any thoughts on why C/I is down?

daemons/naggerd/nagger.go
jacobhaven: The mail "spooler" could keep track of all `(uid, msgid, ntime)` and only send unique tuples. This will also be a pretty loud failure, as we'll log the DB error and also log restarting the container.

maxtaco: Hm, this just punts the same problem to the email spooler....

jacobhaven: If we're truly worried about duplicate/dropped reminders, the spooler is in the best position to know which reminders have actually been sent out.

jacobhaven:
*Fixed the C/I issue* (we were passing `flag.ExitOnError` which didn't behave well in the tests)

daemons/naggerd/nagger_test.go
```diff
@@ -0,0 +1,85 @@
...
+	test.AddReminder(n.sm, time.Millisecond)
+	test.AddReminder(n.sm, 2*time.Second+time.Millisecond)
+	assert.Equal(t, 0, len(remind.rms), "no reminders should be sent yet")
+
+	fc.Advance(time.Second)
+	assert.Equal(t, 1, len(remind.rms), "1 reminders should have been sent")
+	fc.Advance(time.Second)
+	assert.Equal(t, 1, len(remind.rms), "1 reminders should have been sent")
+	fc.Advance(time.Second)
+	assert.Equal(t, 2, len(remind.rms), "2 reminders should have been sent")
```
maxtaco: can you further test that the content of the reminders is correct?  What if we have two different Creations with different reminder intervals.  Can you test that they arrive in the right order?  Just testing the number of reminders sent could be masking bugs.

maxtaco:
OK 👍 whenever C/I goes green

PR #50, opened 2016-04-22
implement logging in gregord

oconnor663:
Make shared parts of the code take logger objects as arguments. The main
function instantiates a go-logging backend, with optional support for
syslog. The client will eventually pass in a client logging
implementation of its own.

This PR will probably replace https://github.com/keybase/gregor/pull/48, which didn't take a contextified approach to the shared code that client will be using. Depending on `rpc.LogOutput` was the easiest thing here, but we can also have gregor reach further and depend on `client/go/logger` if we want. (We might have to make a much larger wrapper object in that case, because the client interface has a couple dozen methods.)

I also switched out `logrus` for `go-logging`, because once we're passing contextified objects around there's essentially no difference between them at all, and the latter gives us a little more control over the multiple backends.

r? @jacobhaven @maxtaco

### DIFF
--- a/gregord/consumer.go
+++ b/gregord/consumer.go
@@ -3,26 +3,28 @@ package main
 import (
 	"database/sql"
 	"errors"
-	"log"
 	"net/url"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/keybase/gregor"
 	"github.com/keybase/gregor/protocol/gregor1"
 	"github.com/keybase/gregor/storage"
+
+	rpc "github.com/keybase/go-framed-msgpack-rpc"
 )
 
 type consumer struct {
 	db *sql.DB
 	sm gregor.StateMachine
 	gregor1.IncomingInterface
+	log rpc.LogOutput
 }
 
-func newConsumer(dsn *url.URL) (*consumer, error) {
+func newConsumer(dsn *url.URL, log rpc.LogOutput) (*consumer, error) {
 	if dsn == nil {
 		return nil, errors.New("nil mysql dsn provided to newConsumer")
 	}
-	c := &consumer{}
+	c := &consumer{log: log}
 	if err := c.setup(dsn); err != nil {
 		return nil, err
 	}
@@ -47,7 +49,7 @@ func (c *consumer) shutdown() {
 
 func (c *consumer) createDB(dsn *url.URL) error {
 	dsn = storage.ForceParseTime(dsn)
-	log.Printf("opening mysql connection to %s", dsn)
+	c.log.Info("opening mysql connection to %s", dsn)
 	db, err := sql.Open("mysql", dsn.String())
 	if err != nil {
 		return err
--- a/gregord/consumer.go
+++ b/gregord/consumer.go
@@ -6,11 +6,10 @@ import (
 	"net/url"
 
 	_ "github.com/go-sql-driver/mysql"
+	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"github.com/keybase/gregor"
 	"github.com/keybase/gregor/protocol/gregor1"
 	"github.com/keybase/gregor/storage"
-
-	rpc "github.com/keybase/go-framed-msgpack-rpc"
 )
 
 type consumer struct {
--- a/gregord/log.go
+++ b/gregord/log.go
@@ -43,7 +43,7 @@ var noColorFormat = `%{level:.4s} %{time:15:04:05.000} %{shortfile} %{message}`
 
 var colorFormat = "%{color}" + noColorFormat + "%{color:reset}"
 
-func newLogger() *logging.Logger {
+func newLogger() rpc.LogOutput {
 	backend := logging.NewLogBackend(os.Stderr, "", 0)
 	var format string
 	if terminal.IsTerminal(int(os.Stdout.Fd())) {
@@ -74,5 +74,5 @@ func newLogger() *logging.Logger {
 
 	logger := logging.MustGetLogger("gregord")
 	logger.SetBackend(logging.MultiLogger(backends...))
-	return logger
+	return &GoLoggingWrapperForRPC{logger}
 }
--- a/gregord/log.go
+++ b/gregord/log.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"fmt"
+	logging "github.com/keybase/go-logging"
+	"golang.org/x/crypto/ssh/terminal"
+	"log/syslog"
+	"os"
+)
+
+// The gregor libraries don't depend on any particular logging implementation.
+// Rather they depend on the go-framed-msgpack-rpc library's LogOutput
+// interface, and they take implementation of that provided by the caller. This
+// allows client code to pass in the client's Logger object, for gregor code
+// running inside the client. Here in the server we implement the LogOutput
+// interface using the go-logging library.
+
+type GoLoggingWrapperForRPC struct {
+	inner *logging.Logger
+}
+
+func (g *GoLoggingWrapperForRPC) Error(s string, args ...interface{}) {
+	g.inner.Errorf(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Warning(s string, args ...interface{}) {
+	g.inner.Warningf(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Info(s string, args ...interface{}) {
+	g.inner.Infof(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Debug(s string, args ...interface{}) {
+	g.inner.Debugf(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Profile(s string, args ...interface{}) {
+	g.inner.Debugf(s, args...)
+}
+
+var noColorFormat = `%{level:.4s} %{time:15:04:05.000} %{shortfile} %{message}`
+
+var colorFormat = "%{color}" + noColorFormat + "%{color:reset}"
+
+func newLogger() *logging.Logger {
+	backend := logging.NewLogBackend(os.Stderr, "", 0)
+	var format string
+	if terminal.IsTerminal(int(os.Stdout.Fd())) {
+		format = colorFormat
+	} else {
+		format = noColorFormat
+	}
+	formattedBackend := logging.NewBackendFormatter(backend, logging.MustStringFormatter(format))
+	backends := []logging.Backend{formattedBackend}
+
+	// Add a syslog backend if the right env vars are present.
+	// TODO: *Remove* the stdout backend in this case?
+	syslogIP := os.Getenv("LOGGLY_PORT_514_UDP_ADDR")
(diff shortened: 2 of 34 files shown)


gregord/vendor/vendor.json
```diff
@@ -66,8 +66,8 @@
 		},
 		{
 			"path": "github.com/keybase/gregor/rpc",
-			"revision": "981e40adf9373075abd789255a21b2e56ec57ffb",
-			"revisionTime": "2016-04-21T10:43:38-07:00"
+			"revision": "b08edbb2cb468f1a4cc43d473f868d870e42e580",
```
oconnor663: This hash is probably going to be meaningless when this lands, and I need to remember to update it.

gregord/main.go
```diff
@@ -56,3 +63,61 @@ func (m mockAuth) RevokeSessionIDs(_ context.Context, sessionIDs []gregor1.Sessi
 }
 
 var _ gregor1.AuthInterface = mockAuth{}
+
+type GoLoggingWrapperForRPC struct {
```
jacobhaven: I'd move this to a `rpc/log_wrapper.go`

gregord/vendor/github.com/keybase/gregor/rpc/server.go
```diff
@@ -82,11 +83,13 @@ type Server struct {
...
 	// code and wait for something bad to happen.
 	useDeadlocker bool
+
+	log rpc.LogOutput
 }
 
 // NewServer creates a Server.  You must call ListenLoop(...) and Serve(...)
 // for it to be functional.
-func NewServer() *Server {
+func NewServer(log rpc.LogOutput) *Server {
```
jacobhaven: I'd take a `*logging.Logger` and wrap it with `GoLoggingWrapperForRPC` from `log_wrapper.go`.

storage/client.go
```diff
@@ -22,15 +23,17 @@ type Client struct {
 	sm       gregor.StateMachine
 	storage  LocalStorageEngine
 	incoming gregor1.IncomingInterface
+	log      rpc.LogOutput
```
jacobhaven: No reason to add `go-framed-msgpack-rpc` everywhere. Instead pass around `*logging.logger`

gregord/main.go
oconnor663: Part of my thinking here was to have only the `main` package be aware that go-logging is what we're using to implement this for gregord.

storage/client.go
oconnor663: @jacobhaven and I discussed this in Slack, but I'll summarize: `logging.Logger` is unfortunately a `struct` and not an `interface`. If we depend on it directly, we won't be able to pass the client's logger implementation in here, even in a wrapped way. Unfortunately logrus has a similar problem.

storage/client.go
```diff
@@ -3,12 +3,13 @@ package storage
...
 	"bytes"
 	"errors"
-	"log"
 	"time"
 
 	"github.com/keybase/gregor"
 	"github.com/keybase/gregor/protocol/gregor1"
 	"golang.org/x/net/context"
+
+	rpc "github.com/keybase/go-framed-msgpack-rpc"
```
jacobhaven: It's a bit annoying to add `go-framed-msgpack-rpc` dependency everywhere.

gregord/main.go
jacobhaven: Sure. `gregord/log_wrapper.go`, then.

gregord/main.go
```diff
@@ -12,38 +13,44 @@ import (
 	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"github.com/keybase/gregor/protocol/gregor1"
 	grpc "github.com/keybase/gregor/rpc"
+
+	logging "github.com/keybase/go-logging"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 func main() {
+	logger := newLogger()
```
jacobhaven: Why not have `newLogger` just return an `rpc.LogOutput` (`* GoLoggingWrapperForRPC`)?

gregord/main.go
```diff
@@ -12,38 +13,44 @@ import (
 	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"github.com/keybase/gregor/protocol/gregor1"
 	grpc "github.com/keybase/gregor/rpc"
+
```
jacobhaven: Remove this extra newline (and rerun goimports)

storage/client.go
oconnor663: Agreed, though I think the only good alternative is adding `github.com/client/keybase/go/logger`. And unfortunately the the interface that logger exposes has way more methods, so it's more annoying to implement.

gregord/main.go
oconnor663: That was my first thought, but `rpc.LogOutput` doesn't expose `Fatal()`. So if we want to use it consistently here, we'll need to replace the `Fatal()` calls below with probably `Error()` + `os.Exit()`. I'm ok with that if you prefer it.

aalness:
>I also switched out logrus for go-logging, because once we're passing contextified objects around there's essentially no difference between them at all

I'm not sure _exactly_ what is meant by contextified objects so maybe what I'm about to say is already being taken into consideration, but one important difference in the two is that one is using a format-string interface while the other promotes a more a structured interface with [a notion of fields](https://github.com/Sirupsen/logrus#fields). Pairing this with JSON can give you fairly powerful capabilities. For example, in KBFS we can quickly see all events that happened for a given user, or device, or folder, or device x folder, etc. across all server instances and with types even build on-demand graphs of things like RPC response times per message type. I don't know that this capability is critical to get gregor going nor should it be a blocker but it might be easier to start out thinking this way vs. retrofitting to achieve it later on.

oconnor663:
@aalness oops I guess contextified means like three different things in our code :p What I meant here was the difference between logging with some global object, and logging with some local object that's passed in by callers. We need to be doing the latter to be compatible with client logging. The plan is for the client to pass it's usual `G.Log` object into the gregor libraries it calls.

For that reason, I don't think we want to expose the structured fields logging at first. The client `Logger` interface doesn't expose anything like that right now. Maybe in the future it would be a good idea to simplify/improve what we have under `client/go/logger` and proclaim that This Is The Logging Implementation For All Of Keybase, and at that point we could add structured fields everywhere?

oconnor663:
@maxtaco @jacobhaven thumbs?

gregord/consumer.go
```diff
@@ -3,26 +3,28 @@ package main
...
 	"database/sql"
 	"errors"
-	"log"
 	"net/url"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/keybase/gregor"
 	"github.com/keybase/gregor/protocol/gregor1"
 	"github.com/keybase/gregor/storage"
+
```
jacobhaven: Remove newline, rerun goimports.

gregord/log.go
```diff
@@ -0,0 +1,74 @@
...
+func (g *GoLoggingWrapperForRPC) Info(s string, args ...interface{}) {
+	g.inner.Infof(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Debug(s string, args ...interface{}) {
+	g.inner.Debugf(s, args...)
+}
+func (g *GoLoggingWrapperForRPC) Profile(s string, args ...interface{}) {
+	g.inner.Debugf(s, args...)
+}
+
```
jacobhaven: Add type assertion
```Go
var _ rpc.LogOutput = (*GoLoggingWrapperForRPC)(nil)
```

gregord/log.go
```diff
@@ -0,0 +1,74 @@
...
+var noColorFormat = `%{level:.4s} %{time:15:04:05.000} %{shortfile} %{message}`
+
+var colorFormat = "%{color}" + noColorFormat + "%{color:reset}"
+
+func newLogger() *logging.Logger {
```
jacobhaven: I'd return an `rpc.LogOutput` for consistency with the rest of the logging code. That is:
```Go
func newLogger() rpc.LogOutput {
	// ...
	return &GoLoggingWrapperForRPC{logger}
}
```

gregord/main.go
```diff
@@ -15,35 +14,38 @@ import (
 )
 
 func main() {
+	logger := newLogger()
```
jacobhaven: And merge `logger` and `rpcLogger` into a single `log` variable, to match naming conventions elsewhere.

rpc/server.go
```diff
@@ -3,14 +3,15 @@ package rpc
...
 	"errors"
-	"log"
 	"net"
 	"time"
 
 	"github.com/jonboulle/clockwork"
 	"github.com/keybase/gregor"
 	"github.com/keybase/gregor/protocol/gregor1"
 	"golang.org/x/net/context"
+
```
jacobhaven: Remove newline here, rerun goimports.

rpc/server.go
```diff
@@ -82,11 +83,13 @@ type Server struct {
...
 	// code and wait for something bad to happen.
 	useDeadlocker bool
+
+	log rpc.LogOutput
 }
 
 // NewServer creates a Server.  You must call ListenLoop(...) and Serve(...)
 // for it to be functional.
-func NewServer() *Server {
+func NewServer(log rpc.LogOutput) *Server {
```
jacobhaven: Maybe we should add to these:
```Go
if log == nil {
	log = rpc.SimpleLogOutput{}
}
```

(and then pass `nil` in the tests)

gregord/main.go
oconnor663: I still need both types visible to support the `Fatal` calls that we're using below, unless you think we should get rid of those. Can definitely `s/logger/log/` though.

gregord/log.go
oconnor663: Same issue as [below](https://github.com/keybase/gregor/pull/50#discussion_r60950536).

rpc/server.go
oconnor663: I think that makes sense for the underlying `go-framed-msgpack-rpc` lib, which does accept nil in these places, but I think it makes less sense for us in gregor. We know that all of our gregor callers really do want to pass in an explicit logger, either for log forwarding on the client or for syslog on the server. I'd prefer to disallow nil if we can, to prevent callers from forgetting that.

gregord/main.go
jacobhaven: I think we should just add `Fatal` to `rpc.LogOutput`

jacobhaven:
Using `log rpc.LogOutput` in `gregord/main.go` is my last comment. Otherwise 👍

gregord/main.go
```diff
@@ -13,38 +13,41 @@ import (
...
 
 func main() {
-	logger := newLogger()
-	rpcLogger := &GoLoggingWrapperForRPC{logger}
+	log := newLogger()
 
 	opts, err := ParseOptions(os.Args)
 	if err != nil {
-		logger.Fatal(err)
+		log.Error(err.Error())
```
jacobhaven: Should probably be `log.Error("%v", err)`

PR #52, opened 2016-04-28
Check if UID is nil in checkMessageAuth

jacobhaven:
r? @maxtaco


maxtaco:
By my read, this is a security hole. It means I'm authenticated by default if it's a nil UID.

jacobhaven:
Correct. Not a security hole though. A message with `nil` UID should be readable by anyone.

maxtaco:
Let's be conservative here and reject any messages with a nil UID as unauthorized. We can loosen the restriction later if we decide we need it (I don't see any reasons now why we'd need it).

maxtaco:
👍

PR #53, opened 2016-04-28
improve the filenames in gregor log output

oconnor663:
Previously we were wrapping the logger object but not setting
ExtraCalldepth, so all our log lines uselessly referred to the log.go
wrappers.

r? @jacobhaven @maxtaco

### DIFF
--- a/daemons/log.go
+++ b/daemons/log.go
@@ -74,5 +74,6 @@ func NewLogger() rpc.LogOutput {
 
 	logger := logging.MustGetLogger("gregord")
 	logger.SetBackend(logging.MultiLogger(backends...))
+	logger.ExtraCalldepth = 1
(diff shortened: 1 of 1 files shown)


oconnor663:
(Had some code in the first version of this PR that set `longfile` instead of `shortfile`. I've removed that.)

jacobhaven:
👍 
I was wondering about those `log.go` statements...

oconnor663:
There will still be log.go statements, but they will now semi-correctly refer to rpc/log.go. Unfortunately that file wraps the logger _again_, but there's no interface for it to bump the call depth, so rpc still has the problem that all gregor code had before this diff.

oconnor663:
In other news, while I agree that macros are an abused feature in general, logging statements are a very good use case for them :p

PR #57, opened 2016-04-29
Cleanup *sql.DB handling to avoid intermittent C/I failures

jacobhaven:
r? @oconnor663 @maxtaco


daemons/gregord/flow_test.go
```diff
@@ -81,34 +81,14 @@ func maybeSleep() {
...
 	srv.SetEventHandler(e)
 
-	consumer, err := newConsumer(u, rpc.SimpleLogOutput{})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	ms := newMainServer(&opts, srv)
+	consumer := newConsumer(db, rpc.SimpleLogOutput{})
+	ms := newMainServer(&Options{BindAddress: "127.0.0.1:0"}, srv)
```
jacobhaven: `mainServer` only uses `opts.BindAddress` and `opts.TLSConfig`, so we only need to set those in the test.

daemons/config.go
```diff
@@ -139,11 +140,11 @@ func (v *FMPURIGetter) String() string {
 type DSNGetter struct {
 	S      string
 	S3conf *S3Config
-	val    *url.URL
+	val    string
```
jacobhaven: We don't actually handle DSN's as `*url.URL`s after `storage.ForceParseTime` has been called, so returning them as strings is more convenient for actual use.

maxtaco:
@patrickxb said he'd take a look

daemons/config.go
patrickxb: what's the point of `v.S`?  and shouldn't DSNGetter.String() return v.val and not v.S?

patrickxb: I guess I don't see the utility of the FMPURIGetter, DSNGetter pattern here...can you explain it to me?  Why does `Get()` return an `interface{}`?

jacobhaven: It's not the best interface, but it's what the `flag` package expects. https://golang.org/pkg/flag/#Getter

patrickxb: Oh, I didn't realize these were for flags.  Could you add a comment?

jacobhaven:
I was able to rerun the C/I a number of times until it failed again:
`test_db_sync.go:61: Error 1215: Cannot add foreign key constraint`
It looks like it's trying to connect to the DB and reset the schema but not able to. There might be a bug in the SQL schema.

patrickxb:
Your dsn changes LGTM pending comment on the *Getter types as flags.

patrickxb:
That's a mysterious error...no idea why it would be intermittent...seems like it should fail every time or not at all.

jacobhaven:
I added a one-time DB initialization script to be run by travis before the tests (the tests no longer try to init their own DB). Also renamed `daemons` to just `bin`.

daemons/vendor/github.com/keybase/gregor/storage/test_db_sync.go
```diff
@@ -2,69 +2,71 @@ package storage
...
-	if name == "" {
-		t.Skip("TEST_MYSQL_DSN not set")
-	}
-	// We need to have parseTime=true as one of our DSN paramenters,
-	// so if it's not there, add it on. This way, in sql_scan, we'll get
-	// time.Time values back from MySQL.
-	u, err := url.Parse(name)
-	if err != nil {
-		t.Fatalf("couldn't parse TEST_MYSQL_DSN: %v", err)
+	mysqlDSN := &daemons.DSNGetter{S: os.Getenv("MYSQL_DSN")}
```
jacobhaven: Parse DSNs with the same `DSNGetter` everywhere.

maxtaco:
Took a quick look but once ci is green ok to merge

PR #58, opened 2016-04-29
Convert naggerd to single-use executable for AWS Lambda

jacobhaven:
r? @aalness


bin/nagger/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+var child_process = require('child_process'),
+    go_proc = null,
+    done = console.log.bind(console),
+    fails = 0,
+    current_event = null;
+
+(function new_go_proc() {
+    // pipe stdin/out, blind passthru stderr
+    go_proc = child_process.spawn('./nagger', { stdio: ['pipe', 'pipe', process.stderr] });
```
jacobhaven: This doesn't pass in the necessary environment variables, need to fix that.

bin/nagger/index.js
```diff
@@ -0,0 +1,98 @@
...
+    fails = 0,
+    current_event = null;
+
+(function new_go_proc() {
+    // pipe stdin/out, blind passthru stderr
+    go_proc = child_process.spawn('./nagger', { 
+    	stdio: ['pipe', 'pipe', process.stderr],
+    	env: {
+    		"AWS_REGION": "us-east-1",
+    		"S3_CONFIG_BUCKET": "kbfs-config-dev",
```
jacobhaven: Hard-coded to dev for now. Eventually we should have `make dev` and `make prod` targets.

jacobhaven:
Couldn't get Lambda to deploy successfully.

PR #59, opened 2016-04-30
fix bug @oconnor663 found

maxtaco:
- a null JSON object wasn't getting properly mapped to a nil Go interface
- fix that with explicit conversion

### DIFF
--- a/protocol/gregor1/extras.go
+++ b/protocol/gregor1/extras.go
@@ -304,8 +304,27 @@ func (i *localIncoming) ConsumeMessage(_ context.Context, msg Message) error {
 	return i.sm.ConsumeMessage(msg)
 }
 
+// DeviceID returns the deviceID in a SyncArc, or interface nil
+// (and not gregor1.DeviceID(nil)) if not was specified.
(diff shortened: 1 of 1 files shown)


jacobhaven:
👍

oconnor663:
Thanks!

protocol/gregor1/extras.go
```diff
@@ -304,8 +304,27 @@ func (i *localIncoming) ConsumeMessage(_ context.Context, msg Message) error {
 	return i.sm.ConsumeMessage(msg)
 }
 
+// DeviceID returns the deviceID in a SyncArc, or interface nil
+// (and not gregor1.DeviceID(nil)) if not was specified.
+func (s SyncArg) DeviceID() gregor.DeviceID {
+	if s.Deviceid == nil {
```
oconnor663: Maybe take this opportunity to `s/Deviceid/deviceid` and `s/Uid/uid` for privacy?

maxtaco: Can't do that, since these are serialized fields and won't show up in RPC otherwise!

PR #60, opened 2016-04-30
Remove flaky flag test

jacobhaven:
This test case would error in different ways depending on the environment variables you had set.

r? @maxtaco


maxtaco:
But I'm confused, why does setting MYSQL_DSN fix the bug?

jacobhaven:
`os.Getenv("MYSQL_DSN")` is the default value for the `-mysql-dsn` flag. If its set, you'll get a session server error, if not you'll get a mysql-dsn error. Bad test case.

maxtaco:
OK, I understand. 👍

PR #61, opened 2016-05-02
avoid a value-method-receiver gotcha in SetCTime

oconnor663:
r? @jacobhaven @maxtaco 

I noticed that `protocol/gregor1/extras.go` defines a lot of value receivers, and that made me want to look for cases where the methods there were trying to modify the receiver (because these modifications will happen on copies and get silently dropped). I believe `Metadata.SetCTime` is one of these. This diff makes all of its methods take a pointer instead, and propagates that change out to the other structs that contain it.

Do you guys think this is the right fix, or would it be better to remove `SetCTime` from the `Metadata` interface, and keep the value type?

### DIFF
--- a/protocol/gregor1/extras.go
+++ b/protocol/gregor1/extras.go
@@ -61,27 +61,27 @@ func (d Dismissal) MsgIDsToDismiss() []gregor.MsgID {
 	return ret
 }
 
-func (m Metadata) CTime() time.Time     { return FromTime(m.Ctime_) }
-func (m Metadata) SetCTime(t time.Time) { m.Ctime_ = ToTime(t) }
-func (m Metadata) UID() gregor.UID {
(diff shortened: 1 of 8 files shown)


protocol/gregor1/extras.go
```diff
@@ -61,27 +61,27 @@ func (d Dismissal) MsgIDsToDismiss() []gregor.MsgID {
 	return ret
 }
 
-func (m Metadata) CTime() time.Time     { return FromTime(m.Ctime_) }
-func (m Metadata) SetCTime(t time.Time) { m.Ctime_ = ToTime(t) }
```
oconnor663: If I have this right, this line was a bug.

oconnor663:
Oops, I broke the build. But also Max suggests we might want to change the interface instead of doing this. Will update shortly.

oconnor663:
Closing in favor of https://github.com/keybase/gregor/pull/62.

PR #62, opened 2016-05-02
Fix failed mutation of CTime

maxtaco:
- Let's just make a new object, it's easier.


maxtaco:
@jacobhaven see above.  I think we have another problem which is that this problem ought to have broken tests, and we didn't see it.  So we need some more tests too

oconnor663:
LGTM

PR #63, opened 2016-05-02
fix an impossible cast

oconnor663:
When we implement an interface with by-value methods, the pointer type
implicitly also implements the interface. That means the compiler won't
yell at you if you try to cast back into a pointer, even though the cast
will in practice always fail, because we use values everywhere.

r? @maxtaco @jacobhaven (@patrickxb would you like to be cc'd on gregor PRs too?)

### DIFF
--- a/bin/vendor/github.com/keybase/gregor/protocol/gregor1/extras.go
+++ b/bin/vendor/github.com/keybase/gregor/protocol/gregor1/extras.go
@@ -204,14 +204,14 @@ func (m Message) ToInBandMessage() gregor.InBandMessage {
 	if m.Ibm_ == nil {
 		return nil
 	}
-	return m.Ibm_
+	return *m.Ibm_
 }
 
 func (m Message) ToOutOfBandMessage() gregor.OutOfBandMessage {
 	if m.Oobm_ == nil {
 		return nil
 	}
-	return m.Oobm_
+	return *m.Oobm_
 }
 
 func (r Reminder) Item() gregor.Item     { return r.Item_ }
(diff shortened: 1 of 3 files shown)


maxtaco:
👍 We definitely have some buggy tests to let this bug through!

oconnor663:
Hmm, this fix actually _breaks_ a test:

```
--- FAIL: TestLevelDBClient (0.00s)
        client_test.go:120: local state hash != server state hash
FAIL
exit status 1
FAIL    github.com/keybase/gregor/storage       0.434s
```

oconnor663:
I'm digging into this, but I'm not too familiar with the code. Here's a little more detail after inserting print statements:

```
here I am calling Hash from extras.go:Sync
IM COMPUTING A HASH ON 7 ITEMS
here I am calling Hash from client.go:syncFromTime
IM COMPUTING A HASH ON 4 ITEMS
--- FAIL: TestLevelDBClient (0.00s)
        client_test.go:120: local state hash != server state hash
```

maxtaco:
Well the test was broken in the first place...

maxtaco:
... So likely the sync code is broken and the tests are broken too...

oconnor663:
Meanwhile the same print statements in master give me:

```
IM COMPUTING A HASH ON 7 ITEMS
here I am calling Hash from client.go:syncFromTime
IM COMPUTING A HASH ON 7 ITEMS
IM COMPUTING A HASH ON 7 ITEMS
here I am calling Hash from client.go:syncFromTime
IM COMPUTING A HASH ON 10 ITEMS
IM COMPUTING A HASH ON 7 ITEMS
here I am calling Hash from client.go:syncFromTime
IM COMPUTING A HASH ON 10 ITEMS
[I] Sync failure: local state hash != server state hash
Reseting StateMachine and retrying
IM COMPUTING A HASH ON 7 ITEMS
here I am calling Hash from client.go:syncFromTime
IM COMPUTING A HASH ON 7 ITEMS
PASS
ok      github.com/keybase/gregor/storage       0.005s
```

oconnor663:
Note to self: add a warning in the failed case.

oconnor663:
Ok, the problem was that `ToInBandMessage` and `ToOutOfBandMessage` in extras.go were returning pointers. That meant the cast was correct for the mem_sql implementation, and I "broke" it by expecting a value. The fix I'm pushing now has us use values instead of pointers everywhere.

PR #64, opened 2016-05-05
CORE-2960, expand gregord cmdline opts

mmaxim:
I expect that there will be comments on this one, but I figured I'd put it up to see what you all thought @oconnor663 and @maxtaco


maxtaco:
And/or @patrickxb.  I should be in by noon or so today !!!

bin/gregord/config.go
```diff
@@ -48,6 +50,7 @@ Environment Variables
     -tls-cert or TLS_CERT
     -aws-region or AWS_REGION
     -s3-config-bucket or S3_CONFIG_BUCKET
+    -rpc-debug OR GREGOR_RPC_DEBUG
```
oconnor663: Lowercase `or` here for consistency with above?

mmaxim: Yep, good call!

bin/log.go
```diff
@@ -72,8 +81,54 @@ func NewLogger() rpc.LogOutput {
...
+
+func (r *RPCLogOptions) ShowAddress() bool    { return !r.noAddress }
+func (r *RPCLogOptions) ShowArg() bool        { return r.verboseTrace }
+func (r *RPCLogOptions) ShowResult() bool     { return r.verboseTrace }
+func (r *RPCLogOptions) Profile() bool        { return r.profile }
+func (r *RPCLogOptions) ClientTrace() bool    { return r.clientTrace }
+func (r *RPCLogOptions) ServerTrace() bool    { return r.serverTrace }
+func (r *RPCLogOptions) TransportStart() bool { return r.connectionInfo }
+
+func NewRPCLogOptions(opts string, log rpc.LogOutput) rpc.LogOptions {
```
oconnor663: It looks like a lot of this is duplicated from client. Any way we could move it into `go-framed-msgpack-rpc` or something like that?

bin/gregord/config.go
```diff
@@ -82,6 +85,7 @@ func parseOptions(argv []string, quiet bool) (*Options, error) {
 	fs.StringVar(&tlsCert, "tls-cert", os.Getenv("TLS_CERT"), "file or S3 bucket or raw TLS Cert")
 	fs.Var(sessionServer, "session-server", "host:port of the session server")
 	fs.Var(mysqlDSN, "mysql-dsn", "user:pw@host/dbname for MySQL")
+	fs.StringVar(&options.RPCDebug, "rpc-debug", os.Getenv("GREGOR_RPC_DEBUG"), "RPC debug options")
```
oconnor663: When I run gregor myself, I pretty much always want it in debug mode. Should we default this to `true`? Not sure myself; curious what you guys think.

bin/log.go
mmaxim: I suppose I could add it to `go-framed-msgpack-rpc` as the type `SimpleOptions` if you all think that is the better play here. I could also update the client to use it as well.

bin/gregord/config.go
mmaxim: Seems like it would be something that changes over time, I picked this way to be more consistent with the client.

mmaxim:
@oconnor663 I deleted that copied code and it now uses stuff from `go-framed-msgpack-rpc`

oconnor663:
LGTM

PR #65, opened 2016-05-06
refactor sync, client and protocol

maxtaco:
This is a refactor to try to address the code sprawl I'm seeing in Gregor:

- remove the `consumer` class in `gregord`
- remove all business logic from `extras.go`; it's mainly for grafting methods onto auto-generated protocol stubs.
- factor out `Sync` so we can call it from `gregord` and from client tests
- use appropriate error wrapper/unwrappers in tests
- split out rpc/server and rpc/client


bin/vendor/github.com/keybase/gregor/rpc/sync.go
```diff
@@ -0,0 +1,34 @@
+package rpc
```
patrickxb: Seems weird to me that gregor/bin vendors gregor/[other stuff]...

rpc/client/client.go
```diff
@@ -0,0 +1,126 @@
...
+	if !bytes.Equal(res.Hash, hash) {
+		return errHashMismatch{}
+	}
+
+	return nil
+}
+func (c *Client) Sync() error {
+	if err := c.syncFromTime(c.sm.LatestCTime(c.user, c.device)); err != nil {
+		if _, ok := err.(errHashMismatch); ok {
+			c.log.Info("Sync failure: %v\nReseting StateMachine and retrying", err)
```
patrickxb: Reseting -> Resetting

rpc/server/connection.go
```diff
@@ -107,12 +107,19 @@ func (c *connection) checkUIDAuth(ctx context.Context, uid gregor1.UID) error {
...
-		if err := c.checkUIDAuth(ctx, m.ToOutOfBandMessage().UID().Bytes()); err != nil {
+	}
+	if oobm := m.ToOutOfBandMessage(); oobm != nil {
+		if oobm.UID() == nil {
+			return errors.New("no valid UID in message")
+		}
+		if err := c.checkUIDAuth(ctx, oobm.UID().Bytes()); err != nil {
 			return err
 		}
 	}
```
patrickxb: So a message can have both inband and out-of-band?  If not, isn't there something somewhere else that gets the UID from a message that would simplify this?

bin/vendor/github.com/keybase/gregor/rpc/sync.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Sync is used by both the main server, and functions mocking the main server. So
+// factor it out into its own function here.
+func Sync(sm gregor.StateMachine, log rpc.LogOutput, arg gregor1.SyncArg) (gregor1.SyncResult, error) {
+	var res gregor1.SyncResult
+	msgs, err := sm.InBandMessagesSince(arg.UID(), arg.DeviceID(), arg.CTime())
+	if err != nil {
+		return res, err
+	}
+	state, err := sm.State(arg.UID(), arg.DeviceID(), nil)
```
patrickxb: Correct me if I'm wrong, but it seems like we always call this with nil t: `sm.State(uid, deviceID, nil)` except in tests.  

Perhaps we should have two functions:  `sm.State(uid, devid)` and `sm.StateAtTime(uid, devid, t)`.

patrickxb:
A few questions, but in general LGTM. 

Can you walk me through the big picture changes on GH?

bin/vendor/github.com/keybase/gregor/rpc/sync.go
maxtaco: There's sadly no way around this.  Go vendoring is very imperfect.

bin/vendor/github.com/keybase/gregor/rpc/sync.go
maxtaco: This seems like a reasonable idea, but adds more to the interface, which has grown a lot over time, sadly. I was hoping that it would be very small.

rpc/server/connection.go
maxtaco: no, it can only have one or the other.  hmmm.  when did this happen?

maxtaco: btw, this is something that is lagging due to vendoring lagging.

maxtaco: Actually, this is just for checking mesasge auth (this is in master BTW).  Seems better to check both. More conservative.

patrickxb:
LGTM

PR #67, opened 2016-05-09
address CORE-3035

mmaxim:
@maxtaco 

This one isn't in the current sprint, but I thought this change might be interesting to look at. If you think it is legit enough, I can get a test case together. The other big question is if we want to make that timeout configurable from the command line (or some other way). That would probably have an impact on the test case, since we won't want to sit around for 10 seconds every time we run the tests.

### DIFF
--- a/bin/gregord/config.go
+++ b/bin/gregord/config.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"time"
 
 	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"github.com/keybase/gregor/bin"
@@ -21,7 +22,7 @@ type Options struct {
 	TLSConfig        *tls.Config
 	MockAuth         bool
 	RPCDebug         string
-	BroadcastTimeout int
+	BroadcastTimeout time.Duration
 }
 
 const usageStr = `Usage:
@@ -87,7 +88,8 @@ func parseOptions(argv []string, quiet bool) (*Options, error) {
 	fs.Var(sessionServer, "session-server", "host:port of the session server")
 	fs.Var(mysqlDSN, "mysql-dsn", "user:pw@host/dbname for MySQL")
 	fs.StringVar(&options.RPCDebug, "rpc-debug", os.Getenv("GREGOR_RPC_DEBUG"), "RPC debug options")
-	fs.IntVar(&options.BroadcastTimeout, "broadcast-timeout", 10000, "Timeout on client broadcasts")
+	fs.DurationVar(&options.BroadcastTimeout, "broadcast-timeout", 10000*time.Millisecond,
+		"Timeout on client broadcasts")
 
 	if err := fs.Parse(argv[1:]); err != nil {
 		return nil, bin.BadUsage(err.Error())
--- a/bin/gregord/config.go
+++ b/bin/gregord/config.go
@@ -14,13 +14,14 @@ import (
 )
 
 type Options struct {
-	SessionServer *rpc.FMPURI
-	BindAddress   string
-	MysqlDSN      string
-	Debug         bool
-	TLSConfig     *tls.Config
-	MockAuth      bool
-	RPCDebug      string
+	SessionServer    *rpc.FMPURI
+	BindAddress      string
+	MysqlDSN         string
+	Debug            bool
+	TLSConfig        *tls.Config
+	MockAuth         bool
+	RPCDebug         string
+	BroadcastTimeout int
 }
 
 const usageStr = `Usage:
@@ -86,6 +87,7 @@ func parseOptions(argv []string, quiet bool) (*Options, error) {
 	fs.Var(sessionServer, "session-server", "host:port of the session server")
 	fs.Var(mysqlDSN, "mysql-dsn", "user:pw@host/dbname for MySQL")
 	fs.StringVar(&options.RPCDebug, "rpc-debug", os.Getenv("GREGOR_RPC_DEBUG"), "RPC debug options")
+	fs.IntVar(&options.BroadcastTimeout, "broadcast-timeout", 10000, "Timeout on client broadcasts")
 
 	if err := fs.Parse(argv[1:]); err != nil {
 		return nil, bin.BadUsage(err.Error())
--- a/bin/vendor/github.com/keybase/gregor/rpc/server/main.go
+++ b/bin/vendor/github.com/keybase/gregor/rpc/server/main.go
@@ -88,23 +88,28 @@ type Server struct {
 	useDeadlocker bool
 
 	log rpc.LogOutput
+
+	// The amount of time a perUIDServer should wait on a BroadcastMessage
+	// response (in MS)
+	broadcastTimeout int
 }
 
 // NewServer creates a Server.  You must call ListenLoop(...) and Serve(...)
 // for it to be functional.
-func NewServer(log rpc.LogOutput) *Server {
+func NewServer(log rpc.LogOutput, broadcastTimeout int) *Server {
 	s := &Server{
-		clock:           clockwork.NewRealClock(),
-		users:           make(map[string]*perUIDServer),
-		lastConns:       make(map[string]connectionID),
-		newConnectionCh: make(chan *connection),
-		statsCh:         make(chan chan *Stats, 1),
-		syncCh:          make(chan syncArgs),
-		consumeCh:       make(chan messageArgs),
-		broadcastCh:     make(chan messageArgs),
-		closeCh:         make(chan struct{}),
-		confirmCh:       make(chan confirmUIDShutdownArgs),
-		log:             log,
+		clock:            clockwork.NewRealClock(),
+		users:            make(map[string]*perUIDServer),
+		lastConns:        make(map[string]connectionID),
+		newConnectionCh:  make(chan *connection),
+		statsCh:          make(chan chan *Stats, 1),
+		syncCh:           make(chan syncArgs),
+		consumeCh:        make(chan messageArgs),
+		broadcastCh:      make(chan messageArgs),
+		closeCh:          make(chan struct{}),
+		confirmCh:        make(chan confirmUIDShutdownArgs),
+		log:              log,
+		broadcastTimeout: broadcastTimeout,
 	}
 
 	return s
@@ -162,7 +167,7 @@ func (s *Server) addUIDConnection(c *connection) error {
 	}
 
 	if usrv == nil {
-		usrv = newPerUIDServer(res.Uid, s.confirmCh, s.closeCh, s.events, s.log)
+		usrv = newPerUIDServer(res.Uid, s.confirmCh, s.closeCh, s.events, s.log, s.broadcastTimeout)
 		if err := s.setPerUIDServer(res.Uid, usrv); err != nil {
 			return err
 		}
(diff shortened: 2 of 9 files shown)


bin/vendor/github.com/keybase/gregor/rpc/server/user_server.go
```diff
@@ -112,13 +114,22 @@ func (s *perUIDServer) broadcast(a messageArgs) {
...
+		//     of a super slow/buggy  client never getting back to us
+		// 2.) Spawn each call into it's own goroutine so a slow device doesn't
+		//     prevent faster devices from getting these messages timely.
+		go func() {
+			ctx, cancel := context.WithTimeout(a.c, 10000*time.Millisecond)
+			defer cancel()
+			if err := oc.BroadcastMessage(ctx, a.m); err != nil {
+				s.log.Info("[connection %d]: %s", id, err)
+
+				if s.isConnDown(err) {
```
maxtaco: I don't think you can do this safely.  I think you'll get race conditions.

bin/vendor/github.com/keybase/gregor/rpc/server/user_server.go
```diff
@@ -112,13 +114,22 @@ func (s *perUIDServer) broadcast(a messageArgs) {
...
+					s.removeConnection(conn, id)
```
maxtaco: This call isn't go-routine safe.

maxtaco: Suggest accumulating the problem children into a go-routine-safe accumulator (like a channel), and them boot them out at the end.

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -271,7 +271,9 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
 	if err := <-retCh; err != nil {
 		return err
 	}
-	s.broadcastCh <- args
+	// Spawn this off so we don't block the main thread if the per UID
+	// server is not going to respond immediatley
+	go func() { s.broadcastCh <- args }()
```
maxtaco: I'm not sure, does `go s.broadcastCh <- args` work?

mmaxim: Doesn't seem to :(

mmaxim:
I have a new attempt up that I think is much better, let me know!

mmaxim:
@oconnor663 take a look!

maxtaco:
👍 I think this looks great, but it's subtle enough we should put a second set of eyes on it!

rpc/server/main.go
```diff
@@ -271,7 +271,9 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
 	if err := <-retCh; err != nil {
 		return err
 	}
-	s.broadcastCh <- args
+	// Spawn this off so we don't block the main thread if the per UID
+	// server is not going to respond immediatley
+	go func() { s.broadcastCh <- args }()
```
patrickxb: I don't quite understand the need for this.  The PerUIDServer's sendBroadcastCh is buffered so this won't block.  If 1 isn't big enough, we could just increase the buffer size.

patrickxb:
Did you see my question?

rpc/server/main.go
maxtaco: We hit a bug here in testing.  If one of the downstream clients didn't reply to an RPC, the whole gregor process locked up.  It was somewhat subtle, but it was a chain of channels that were getting blocked, and the main loop never returned to the select the next time through the loop.

maxtaco: Increasing the buffer size can't work, but it will just delay the problem.  If we have N clients that don't reply to their RPCs (because they have bad networks, let's say), then only a buffer of size N+1 is big enough. So there's no good N to pick.

patrickxb: Did you try making this channel (s.broadcastCh) buffered?

mmaxim: Making it buffered would have cleared the test case, but in the worst case scenario it's impossible to pick the right buffer size, you just get closer to the right one. Spawning it into a thread solves it for all cases, but I'm curious if there is a downside you are thinking about?

rpc/server/user_server.go
```diff
@@ -112,13 +117,36 @@ func (s *perUIDServer) broadcast(a messageArgs) {
...
+	}
+
+	// Wait on all the calls
+	go func() {
+		wg.Wait()
+		close(errCh)
+	}()
+
+	// Clean up all the connections we determined are dead
+	for ca := range errCh {
```
patrickxb: Since this will wait until errCh is closed, seems to me like if there's a connection that times out that this function will take 10s to complete, which means the serve() loop above will be hanging for 10s.

oconnor663: Could we add in some comments about how this is blocking on line 144, and how line 144 is blocking on all the line 127's :)

mmaxim: That is true, but that was the case beforehand as well (although before it was possible to wait forever). I figured it was ok as long as it was in the UID server, as opposed to the main server loop. (this is additional motivation for the change above, a user could have some buggy device that takes like 10 seconds every time). I'm also going to make the timeout configurable, so we could lower it.

maxtaco: I agree with this approach, I think it's the simplest / more-channel-y way to do it, with the downside pointed out above, but with the upside that it's simple.  And as @mmaxim says, it's only using the one user, once, with the bad device.

rpc/server/main.go
patrickxb: The downside would just be that it is unbounded and there is no control over these goroutines.  If you pick an N, at least you know what N is and can set up something to handle the case of the buffer being full (alert, drop messages, send somewhere else, etc.).

rpc/server/user_server.go
patrickxb: One other question:  shouldn't the RPC library handle timeouts?

rpc/server/main.go
maxtaco: That's a good idea, we can put a buffer of ~1000 let's say on this send, and then do the send in a select, make a big ugly warning, and then say that user has a blocked device in the logs.  This way, a bad user can only hose the bad user.

At a deeper level though we might not be solving any problems, since there's a likely a go-routine per RPC call/reply in the library.

rpc/server/user_server.go
maxtaco: The plan I believe is to handle the timeouts via the Context method above...

rpc/server/main.go
patrickxb: After the buffer is full, we wouldn't be making any more RPC calls, so that would bound the goroutines.

mmaxim: Ok will make the change!

mmaxim:
@patrickxb @maxtaco

bin/gregord/config.go
```diff
@@ -14,13 +14,14 @@ import (
...
-	MockAuth      bool
-	RPCDebug      string
+	SessionServer    *rpc.FMPURI
+	BindAddress      string
+	MysqlDSN         string
+	Debug            bool
+	TLSConfig        *tls.Config
+	MockAuth         bool
+	RPCDebug         string
+	BroadcastTimeout int
```
patrickxb: int -> time.Duration

bin/gregord/config.go
```diff
@@ -86,6 +87,7 @@ func parseOptions(argv []string, quiet bool) (*Options, error) {
 	fs.Var(sessionServer, "session-server", "host:port of the session server")
 	fs.Var(mysqlDSN, "mysql-dsn", "user:pw@host/dbname for MySQL")
 	fs.StringVar(&options.RPCDebug, "rpc-debug", os.Getenv("GREGOR_RPC_DEBUG"), "RPC debug options")
+	fs.IntVar(&options.BroadcastTimeout, "broadcast-timeout", 10000, "Timeout on client broadcasts")
```
patrickxb: fs.IntVar -> fs.DurationVar

rpc/server/server_test.go
```diff
@@ -157,6 +161,33 @@ func (c *client) BroadcastMessage(ctx context.Context, m gregor1.Message) error
 	return nil
 }
 
+type slowClient struct {
+	cli          *client
+	responseTime int
```
patrickxb: int -> time.Duration

patrickxb:
One question:  the tests have `time.Sleep(...)` calls.  Did you try the clockwork.NewFakeClock()?  We've had some issues with `time.Sleep` in tests passing locally and failing in CI.

patrickxb:
Also, I made some comments about using `time.Duration` instead of `int`.  I think it's easier to understand `10000 * time.Millisecond` as opposed to `10000`.

maxtaco:
We might be SOL since the timeout mechanism is threaded all the way through to `net.Context`. We could redo *that* with clockwork.  It might actually be worth it.

patrickxb:
Yeah, I'm ok with time.Sleep, I just wanted to make sure fake clock was considered.

rpc/server/main.go
```diff
@@ -253,7 +258,15 @@ func (s *Server) sendBroadcast(c context.Context, m gregor1.Message) error {
...
 		return nil
 	}
-	srv.sendBroadcastCh <- messageArgs{c, m, nil}
+
+	// If this is going to block, we just drop the broadcast. It means that
+	// the user has a device that is not responding fast enough
+	select {
+	case srv.sendBroadcastCh <- messageArgs{c, m, nil}:
+	default:
+		s.log.Error("user super slow receiving broadcasts, rejecting!")
```
patrickxb: Could you add the UID that is slow to the log message?

rpc/server/server_test.go
```diff
@@ -296,6 +327,123 @@ func TestSuperUser(t *testing.T) {
...
+	go func() {
+		time.Sleep(2 * time.Second)
+		timeout <- true
+	}()
+
+	success := 0
+	select {
+	case <-ev.BcastSent:
+		success++
+	case <-timeout:
```
patrickxb: `case <-time.After(2 * time.Second)` and then you don't need 429-433

rpc/server/user_server.go
```diff
@@ -28,21 +31,24 @@ type perUIDServer struct {
 	events           EventHandler
 
 	log rpc.LogOutput
+
+	broadcastTimeout int // in MS
```
patrickxb: int -> time.Duration

rpc/server/main.go
mmaxim: Yep!

mmaxim:
Also, fake clock was considered, but I figured the real clock here should work out ok. If it becomes a problem I can look into threading that through to `net.Context`

patrickxb:
LGTM!

mmaxim:
Thanks!

bin/gregord/config.go
```diff
@@ -86,6 +88,8 @@ func parseOptions(argv []string, quiet bool) (*Options, error) {
 	fs.Var(sessionServer, "session-server", "host:port of the session server")
 	fs.Var(mysqlDSN, "mysql-dsn", "user:pw@host/dbname for MySQL")
 	fs.StringVar(&options.RPCDebug, "rpc-debug", os.Getenv("GREGOR_RPC_DEBUG"), "RPC debug options")
+	fs.DurationVar(&options.BroadcastTimeout, "broadcast-timeout", 10000*time.Millisecond,
```
jacobhaven: `10*time.Second`

PR #68, opened 2016-05-10
test for the gregord's Sync() RPC 

maxtaco:
test for the gregord's Sync() RPC
    
- just use toy data for now, this test can be beefed up
- expose fake clock and plumb through
- mockAuth is more useful with random UIDs.


patrickxb:
LGTM

PR #69, opened 2016-05-10
address CORE-3020, reconnect to authd

mmaxim:
@maxtaco @patrickxb @oconnor663 here is a fix for CORE-3020, check it out!


maxtaco:
👍 looks good!

PR #72, opened 2016-05-12
PubSub between gregors

patrickxb:
* `srvup` package to maintain alive state in mysql
* `rpc/server` maintains group of alive gregord sibling connections, publishes messages to them

r? @maxtaco

### DIFF
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -81,8 +81,15 @@ func setupPubSub(opts *Options, log *bin.StandardLogger) *srvup.Status {
 		log.Error("%#v", err)
 		os.Exit(3)
 	}
+
 	// start sending heartbeats
-	statusGroup.HeartbeatLoop(opts.BindAddress)
+	externalAddr := opts.BindAddress
+	if len(opts.IncomingAddress) > 0 {
+		// only use opts.IncomingAddress if it is set
+		externalAddr = opts.IncomingAddress
+	}
+	log.Debug("starting heartbeat loop for address %s", externalAddr)
+	statusGroup.HeartbeatLoop(externalAddr)
 
 	if len(alive) > 0 {
 		// there are other gregors up
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/keybase/gregor/bin"
 	"github.com/keybase/gregor/protocol/gregor1"
 	server "github.com/keybase/gregor/rpc/server"
+	"github.com/keybase/gregor/srvup"
 	"github.com/keybase/gregor/storage"
 )
 
@@ -52,6 +53,15 @@ func main() {
 		log.Info("DB close on clean shutdown")
 		db.Close()
 	}()
+
+	mstore, err := srvup.NewStorageMysql(opts.MysqlDSN, log)
+	if err != nil {
+		log.Error("%#v", err)
+		os.Exit(3)
+	}
+	status := srvup.New("gregord", opts.HeartbeatInterval, opts.AliveThreshold, mstore)
+	defer status.Shutdown()
+	status.HeartbeatLoop(opts.BindAddress)
 
 	sm := storage.NewMySQLEngine(db, gregor1.ObjFactory{})
 	srv.SetStorageStateMachine(sm)
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -60,7 +60,7 @@ func main() {
 		os.Exit(3)
 	}
 	statusGroup := srvup.New("gregord", opts.HeartbeatInterval, opts.AliveThreshold, mstore)
-	defer status.Shutdown()
+	defer statusGroup.Shutdown()
 
 	alive, err := statusGroup.Alive()
 	if err != nil {
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -59,9 +59,29 @@ func main() {
 		log.Error("%#v", err)
 		os.Exit(3)
 	}
-	status := srvup.New("gregord", opts.HeartbeatInterval, opts.AliveThreshold, mstore)
+	statusGroup := srvup.New("gregord", opts.HeartbeatInterval, opts.AliveThreshold, mstore)
 	defer status.Shutdown()
-	status.HeartbeatLoop(opts.BindAddress)
+
+	alive, err := statusGroup.Alive()
+	if err != nil {
+		// bad enough to quit:
+		log.Error("%#v", err)
+		os.Exit(3)
+	}
+	// start sending heartbeats
+	statusGroup.HeartbeatLoop(opts.BindAddress)
+
+	if len(alive) > 0 {
+		// there are other gregors up
+		log.Debug("existing gregord servers: %v", alive)
+		log.Debug("sleeping for alive threshold (%s) before starting server", opts.AliveThreshold)
+		time.Sleep(opts.AliveThreshold)
+		log.Debug("sleep complete, proceeding with server initialization")
+	} else {
+		log.Debug("self is first gregord server alive, proceeding directly to server initialization")
+	}
+
+	srv.SetStatusGroup(statusGroup)
 
 	sm := storage.NewMySQLEngine(db, gregor1.ObjFactory{})
 	srv.SetStorageStateMachine(sm)
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -28,7 +28,7 @@ func main() {
 	rpcopts := rpc.NewStandardLogOptions(opts.RPCDebug, log)
 	log.Configure(opts.Debug)
 	log.Debug("Options Parsed. Creating server...")
-	srv := server.NewServer(log, opts.BroadcastTimeout, opts.PublishBufferSize, opts.NumPublishers, opts.StorageHandlers, opts.StorageQueueSize)
+	srv := server.NewServer(log, opts.BroadcastTimeout, opts.PublishBufferSize, opts.NumPublishers, opts.StorageHandlers, opts.StorageQueueSize, opts.PublishTimeout)
 
 	if opts.MockAuth {
 		srv.SetAuthenticator(newMockAuth())
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/keybase/gregor/bin"
 	"github.com/keybase/gregor/protocol/gregor1"
 	server "github.com/keybase/gregor/rpc/server"
+	"github.com/keybase/gregor/rpc/transport"
 	"github.com/keybase/gregor/srvup"
 	"github.com/keybase/gregor/storage"
 )
@@ -37,6 +38,12 @@ func main() {
 		sc := server.NewSessionCacherFromURI(opts.SessionServer, clockwork.NewRealClock(),
 			10*time.Minute, log, rpcopts)
 		defer sc.Close()
+		trans := transport.NewConnTransport(log, rpcopts, opts.SessionServer)
+		handler := NewAuthdHandler(log)
+
+		log.Debug("Connecting to session server %s", opts.SessionServer.String())
+		rpc.NewConnectionWithTransport(&handler, trans, keybase1.ErrorUnwrapper{},
+			true, keybase1.WrapError, log, nil)
 
 		log.Debug("Setting authenticator")
 		srv.SetAuthenticator(sc)
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -54,13 +54,27 @@ func main() {
 		db.Close()
 	}()
 
+	statusGroup := setupPubSub(opts, log)
+	defer statusGroup.Shutdown()
+
+	srv.SetStatusGroup(statusGroup)
+
+	sm := storage.NewMySQLEngine(db, gregor1.ObjFactory{})
+	srv.SetStorageStateMachine(sm)
+	go srv.Serve()
+
+	log.Debug("Calling mainServer.listenAndServe()")
+	log.Error("%#v", newMainServer(opts, srv).listenAndServe())
+	os.Exit(4)
+}
+
+func setupPubSub(opts *Options, log *bin.StandardLogger) *srvup.Status {
 	mstore, err := srvup.NewStorageMysql(opts.MysqlDSN, log)
 	if err != nil {
 		log.Error("%#v", err)
 		os.Exit(3)
 	}
 	statusGroup := srvup.New("gregord", opts.HeartbeatInterval, opts.AliveThreshold, mstore)
-	defer statusGroup.Shutdown()
 
 	alive, err := statusGroup.Alive()
 	if err != nil {
@@ -81,13 +95,6 @@ func main() {
 		log.Debug("self is first gregord server alive, proceeding directly to server initialization")
 	}
 
-	srv.SetStatusGroup(statusGroup)
+	return statusGroup
 
-	sm := storage.NewMySQLEngine(db, gregor1.ObjFactory{})
-	srv.SetStorageStateMachine(sm)
-	go srv.Serve()
-
-	log.Debug("Calling mainServer.listenAndServe()")
-	log.Error("%#v", newMainServer(opts, srv).listenAndServe())
-	os.Exit(4)
 }
--- a/bin/gregord/main.go
+++ b/bin/gregord/main.go
@@ -28,7 +28,15 @@ func main() {
 	rpcopts := rpc.NewStandardLogOptions(opts.RPCDebug, log)
 	log.Configure(opts.Debug)
 	log.Debug("Options Parsed. Creating server...")
-	srv := server.NewServer(log, opts.BroadcastTimeout, opts.PublishBufferSize, opts.NumPublishers, opts.StorageHandlers, opts.StorageQueueSize, opts.PublishTimeout)
+	srvopts := server.ServerOpts{
+		BroadcastTimeout: opts.BroadcastTimeout,
+		PublishChSize:    opts.PublishBufferSize,
+		NumPublishers:    opts.NumPublishers,
+		PublishTimeout:   opts.PublishTimeout,
+		StorageHandlers:  opts.StorageHandlers,
+		StorageQueueSize: opts.StorageQueueSize,
+	}
+	srv := server.NewServer(log, srvopts)
 
 	if opts.MockAuth {
 		srv.SetAuthenticator(newMockAuth())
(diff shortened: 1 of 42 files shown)


bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -285,7 +307,58 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
 	if err := <-retCh; err != nil {
 		return err
 	}
-	s.broadcastCh <- args
+	s.broadcastCh <- messageArgs{c: c, m: m}
+
+	select {
+	case s.publishCh <- messageArgs{c: c, m: m}:
+	default:
+		s.log.Warning("publishCh full: %d", len(s.publishCh))
```
maxtaco: This is super bad if this happens....

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -285,7 +307,58 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
...
+			return
+		}
+	}
+}
+
+func (s *Server) publish(marg messageArgs) error {
+	s.log.Debug("publish: %+v", marg)
+	s.groupOnce.Do(func() {
+		s.log.Debug("creating new aliveGroup")
+		s.group = newAliveGroup(s.statusGroup, s.addr.String(), s.authToken, s.clock, s.closeCh, s.log)
```
maxtaco: Getting an externally routable host/port that we can bind to might be hard in a container.  We likely need to pass it through on the command line to gregor, and figure out a way to plumb it via ECS.

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -329,6 +402,7 @@ func (s *Server) handleNewConnection(c net.Conn) error {
 
 // ListenLoop listens for new connections on net.Listener.
 func (s *Server) ListenLoop(l net.Listener) error {
+	s.addr = l.Addr()
```
maxtaco: see above, this probably isn't going to work...

bin/vendor/github.com/keybase/gregor/srvup/storage_mysql.go
```diff
@@ -0,0 +1,194 @@
...
+// setClock stores a clock for use in testing.  See s.now().
+func (s *StorageMysql) setClock(c clockwork.Clock) {
+	s.clock = c
+}
+
+// UpdateServerStatus implements Storage.UpdateServerStatus.
+func (s *StorageMysql) UpdateServerStatus(group, hostname string) error {
+	var err error
+	if s.update == nil {
+		if s.clock != nil {
```
maxtaco: i had a pretty good story for this exact problem in `storage/sql_sm`.  Is there anyway to use that code here?

srvup/storage_mysql.go
```diff
@@ -0,0 +1,194 @@
...
+func (s *StorageMysql) setClock(c clockwork.Clock) {
+	s.clock = c
+}
+
+// UpdateServerStatus implements Storage.UpdateServerStatus.
+func (s *StorageMysql) UpdateServerStatus(group, hostname string) error {
+	var err error
+	if s.update == nil {
+		if s.clock != nil {
+			s.update, err = s.db.Prepare("INSERT INTO server_status (groupname, hostname, hbtime, ctime) VALUES (?, ?, ?, ?) ON DUPLICATE KEY UPDATE hbtime=?")
```
maxtaco: i had a pretty good story for this exact problem in storage/sql_sm. Is there anyway to use that code here?

maxtaco: Factor our my queryBuilder class?

patrickxb: Yes, I know.  However:

1. I was trying to keep this as independent as possible.  It currently could be used by any service we come up with and be its own separate package.
2. There are only a few queries in this package and I found it easier to read and understand to see the whole query as opposed to the query builder method.

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
patrickxb: It already comes as a parameter, from gregord command line options.

maxtaco: But that bind address might be internal to Docker, and not externally router, since Docker does some NAT'ing or whatnot :(  So we might need to pass 2 addresses, one internal, and one external

maxtaco: (Sorry should have commented on the main file, not the vendored file...)

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
patrickxb: btw, this is a diff on vendoring...you might want to look at the diff of the source.  The vendoring in bin could be out of date.

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
patrickxb: What's the point of an internal to docker bind address for gregor?  Isn't this bind address the address the ELB is using to connect to gregord?

maxtaco: I don't know, that's just how it works. seems like an extra unneeded level of complication

srvup/storage_mysql.go
maxtaco: OK

maxtaco:
Can you `git rm` the `gregor.test` binary that's checked in/

bin/gregord/main.go
```diff
@@ -52,6 +53,35 @@ func main() {
 		db.Close()
 	}()
 
+	mstore, err := srvup.NewStorageMysql(opts.MysqlDSN, log)
```
maxtaco: Can you move this logic to a subroutine to prevent too much clutter in the  main function?

bin/gregord/main.go
```diff
@@ -52,6 +53,35 @@ func main() {
...
+	defer statusGroup.Shutdown()
+
+	alive, err := statusGroup.Alive()
+	if err != nil {
+		// bad enough to quit:
+		log.Error("%#v", err)
+		os.Exit(3)
+	}
+	// start sending heartbeats
+	statusGroup.HeartbeatLoop(opts.BindAddress)
```
maxtaco: Introduce new command-line flag for externally routable incoming IP address.

bin/vendor/github.com/keybase/gregor/protocol/gregor1/extras.go
```diff
@@ -60,8 +60,7 @@ func (d Dismissal) MsgIDsToDismiss() []gregor.MsgID {
 	return ret
 }
 
-func (m Metadata) CTime() time.Time     { return FromTime(m.Ctime_) }
-func (m Metadata) SetCTime(t time.Time) { m.Ctime_ = ToTime(t) }
```
maxtaco: never can quite seem to get rid of this!

patrickxb:
I did already

bin/vendor/github.com/keybase/gregor/protocol/gregor1/extras.go
patrickxb: I didn't make this change...you're in vendor again...

rpc/server/alive_group.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, host := range alive {
+		if _, ok := a.group[host]; !ok {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (a *aliveGroup) update() error {
```
maxtaco: will this function ever be called more than once in parallel?

rpc/server/errors.go
```diff
@@ -1,10 +1,38 @@
...
 func IsSocketClosedError(e error) bool {
 	return strings.HasSuffix(e.Error(), "use of closed network connection")
 }
+
+type hostErr struct {
+	host string
+	err  error
+}
+
+type pubErr struct {
```
maxtaco: can you comment what this is for?

rpc/server/alive_group.go
patrickxb: no.  one goroutine running check/update.

rpc/transport/transport.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	t.stagedTransport = rpc.NewTransport(t.conn, rpc.NewSimpleLogFactory(t.log, t.opts), keybase1.WrapError)
+	return t.stagedTransport, nil
+}
+
+func (t *connTransport) IsConnected() bool {
+	return t.transport != nil && t.transport.IsConnected()
+}
+
+func (t *connTransport) Finalize() {
```
maxtaco: why is this needed?

rpc/server/publish_test.go
```diff
@@ -0,0 +1,74 @@
...
+	select {
+	case <-e2.BcastSent:
+	case <-time.After(2 * time.Second):
+		t.Fatal("timeout waiting for BcastSent on s2")
+	}
+
+	// check that the client connected to s2 received a broadcast
+	if len(cli.broadcasts) != 1 {
+		t.Errorf("client broadcasts received: %d, expected 1", len(cli.broadcasts))
+	}
```
maxtaco: can we check that the pubsub'ed message is sent all the way through to the client listening on the other gregor?

rpc/server/alive_group.go
```diff
@@ -0,0 +1,148 @@
...
+	go a.check()
+	return a
+}
+
+func (a *aliveGroup) Publish(ctx context.Context, msg gregor1.Message) error {
+	a.RLock()
+	defer a.RUnlock()
+	perr := &pubErr{}
+	for host, conn := range a.group {
+		if err := conn.CallConsumePubMessage(ctx, msg); err != nil {
```
maxtaco: does it make sense to do these in serial?

rpc/server/publish_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
```
maxtaco: are there tests for servers going up and down too?  Seems like we want to beef up the tests as much as possible.

rpc/server/publish_test.go
patrickxb: That's what this is.  Message was sent to s1.  cli is connected to s2.  This tests that cli received the message.

rpc/transport/transport.go
patrickxb: I just moved this file from bin/gregord.

rpc/server/publish_test.go
maxtaco: Ok, is it possible to check the msg payload too, not just the existence of the message?

rpc/transport/transport.go
maxtaco: Ok, you and @mmaxim shoudl coordinate, I think this is slightly old...

bin/vendor/github.com/keybase/gregor/rpc/server/alive_group.go
```diff
@@ -0,0 +1,197 @@
...
+			} else {
+				newgroup[host] = newcli
+			}
+		}
+	}
+
+	for host, _ := range a.group {
+		if _, exists := newgroup[host]; !exists {
+			a.log.Debug("gregord on host %q no longer alive, shutting connection down", host)
+			// cli.Shutdown()
```
maxtaco: what's going on here?

patrickxb: urgh, you are in the vendor diff again...let me fix vendoring and repush.  this is different in the repo.

patrickxb: ok, vendoring should be fixed now

rpc/server/main.go
```diff
@@ -106,12 +121,16 @@ func NewServer(log rpc.LogOutput, broadcastTimeout time.Duration) *Server {
 		syncCh:           make(chan syncArgs),
 		consumeCh:        make(chan messageArgs),
 		broadcastCh:      make(chan messageArgs),
+		publishCh:        make(chan messageArgs, 1000),
```
maxtaco: can this be a parameter we can tune?

rpc/server/main.go
```diff
@@ -285,7 +308,60 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
...
+
+// consumePub handles published messages from other gregord
+// servers.  It doesn't store to db or publish.
+func (s *Server) consumePub(c context.Context, m gregor1.Message) error {
+	s.broadcastCh <- messageArgs{c: c, m: m}
+	return nil
+}
+
+func (s *Server) publishSpawn() {
+	for i := 0; i < 10; i++ {
```
maxtaco: this to, can it be tunable?

rpc/transport/transport.go
mmaxim: Yeah in master this is in rpc/server, although this is a fine spot too.

maxtaco:
This is a really good start.  It has interactions with @mmaxim's branch for parallelizing DB writes. This type of code though is pretty hard to get right, so I hesitate to think I've fully understood all of the various failure cases.

rpc/server/main.go
```diff
@@ -285,7 +308,60 @@ func (s *Server) consume(c context.Context, m gregor1.Message) error {
 	if err := <-retCh; err != nil {
 		return err
 	}
-	s.broadcastCh <- args
+	s.broadcastCh <- messageArgs{c: c, m: m}
+
+	select {
+	case s.publishCh <- messageArgs{c: c, m: m}:
+	default:
+		s.log.Warning("publishCh full: %d", len(s.publishCh))
```
mmaxim: Seems like this should be an error?

patrickxb:
Thanks.  I'm sure there will be issues that we won't see until it's running...

maxtaco:
I just have some small comments, and the approach looks solid overall.

patrickxb:
Thanks.

It occurred to me after @mmaxim fix w/ timeouts that these intra-gregor calls also need timeouts...I will add that as well.

bin/gregord/main.go
patrickxb: done.

rpc/server/errors.go
patrickxb: done.

rpc/transport/transport.go
patrickxb: Ok, I don't have a strong preference, just wanted to reuse the code.  Sounds like @mmaxim and I had the same thought in parallel branches.

Can we cleanup after merge?

rpc/server/publish_test.go
patrickxb: done

rpc/server/main.go
patrickxb: Good idea.  I'm returning an error now.

What do you guys think about changing `default` to:

    case <-time.After(500 * time.Millisecond):

in the event of a clogged system, that might allow more messages to get through.  But perhaps having all the consume calls waiting in that scenario is a bad idea.

rpc/server/main.go
patrickxb: done

rpc/server/main.go
patrickxb: done

rpc/server/main.go
patrickxb: Nevermind.  We can just increase the buffer size.

bin/gregord/main.go
patrickxb: done.

rpc/server/alive_group.go
patrickxb: No, I changed to do them concurrently and with a timeout.

patrickxb:
Ok, I had a hell of a rebase, but I think I've addressed all the PR feedback.

@maxtaco @mmaxim PTAL!

patrickxb:
I'm adding some more tests for servers going up and down...

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -293,6 +332,54 @@ func (s *Server) runConsumeMessageMainSequence(c context.Context, m gregor1.Mess
...
 
+	select {
+	case s.publishCh <- messageArgs{c: c, m: m}:
+	default:
+		s.log.Warning("publishCh full: %d", len(s.publishCh))
+		return ErrPublishChannelFull
+	}
+	return nil
+}
+
```
mmaxim: It might make sense to pull this guy into a function called `publishConsumeMessage`. Max wanted this function to really make the whole routine very explicit, which is why I make the broadcast and storage version.

mmaxim: It if wasn't clear, I'm referring to the code above this comment

maxtaco: 👍

patrickxb: got it.  yeah, I saw that as I was rebasing.  Will change now...

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -293,6 +332,54 @@ func (s *Server) runConsumeMessageMainSequence(c context.Context, m gregor1.Mess
...
+// consumePub handles published messages from other gregord
+// servers.  It doesn't store to db or publish.
+func (s *Server) consumePub(c context.Context, m gregor1.Message) error {
+	s.broadcastCh <- messageArgs{c: c, m: m}
+	return nil
+}
+
```
mmaxim: I wonder if we should just pick a name for "publish", either but mixing up "publish" and "pub" might not be the best play. We could think about renaming this to `consumePublish`, since "pub" can kind of map to a bunch of different concepts on first read.

mmaxim: Also I realize this is the vendored code now, it just comes first in the display!

bin/vendor/github.com/keybase/gregor/rpc/server/sibconn.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+
+	return s, nil
+}
+
+func (s *sibConn) CallConsumePubMessage(ctx context.Context, msg gregor1.Message) error {
+	ctx, cancel := context.WithTimeout(ctx, s.timeout)
+	defer cancel()
+	ic := gregor1.IncomingClient{Cli: s.conn.GetClient()}
```
mmaxim: As I discovered after talking to strib, you could just keep around a client on `sibConn`, you just have to grab it in `connect` rather than `OnConnect`

maxtaco:
👍 ready when you are

mmaxim:
Same!

jzila:
This seems to be breaking KBFS docker tests. Here's an example: https://travis-ci.com/keybase/kbfs/jobs/39968094#L798

PR #73, opened 2016-05-12
fix authd reconnecting logic

mmaxim:
@maxtaco this is much simpler now...


maxtaco:
👍 but i might want a little in-person discussion

PR #74, opened 2016-05-12
handle CORE-3028

mmaxim:
@maxtaco @patrickxb 

Add parallelism to accessing the StorageMachine so we don't block the main loop every time we touch it. I'm going to get some tests going, but I wanted to get this in front you guys to make sure you generally agree with the direction I went in.

Thanks!

### DIFF
--- a/bin/vendor/github.com/keybase/gregor/rpc/server/main.go
+++ b/bin/vendor/github.com/keybase/gregor/rpc/server/main.go
@@ -278,7 +278,7 @@ func (s *Server) sendBroadcast(c context.Context, m gregor1.Message) error {
 	return nil
 }
 
-// This function gets called from the connection object that exists for each
+// startSync gets called from the connection object that exists for each
 // client of gregord for a sync call. It is on a different thread than
 // the main Serve loop
 func (s *Server) startSync(c context.Context, arg gregor1.SyncArg) (gregor1.SyncResult, error) {
@@ -288,8 +288,10 @@ func (s *Server) startSync(c context.Context, arg gregor1.SyncArg) (gregor1.Sync
 	return ret.res, ret.err
 }
 
-// Like startSync, this function is called from connection, and is on a
-// different thread than Serve
+// startConsume is the main entry point to the gregor flow described in the
+// architecture documents. It receives a new message, writes it to the StateMachine,
+// and broadcasts it to the other clients. Like startSync, this function is called
+// from connection, and is on a different thread than Serve
 func (s *Server) startConsume(c context.Context, m gregor1.Message) error {
 	retCh := make(chan error)
 	s.storageConsumeMessage(m, retCh)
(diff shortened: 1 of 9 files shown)


mmaxim:
updated to remove `consumeCh` and `syncCh`

bin/vendor/github.com/keybase/gregor/rpc/server/main.go
```diff
@@ -92,24 +92,31 @@ type Server struct {
...
-		broadcastTimeout: broadcastTimeout,
+		clock:             clockwork.NewRealClock(),
+		users:             make(map[string]*perUIDServer),
+		lastConns:         make(map[string]connectionID),
+		newConnectionCh:   make(chan *connection),
+		statsCh:           make(chan chan *Stats, 1),
+		broadcastCh:       make(chan messageArgs),
+		closeCh:           make(chan struct{}),
+		confirmCh:         make(chan confirmUIDShutdownArgs),
+		storageDispatchCh: make(chan storageReq, 10000),
```
maxtaco: reminder to make this tunable

rpc/server/main.go
```diff
@@ -314,6 +320,71 @@ func (s *Server) Serve() error {
...
+	}
+	return nil
+}
+
+// storageDispatchHandler handles pulling requests off the storageDispatchCh queue
+// and routing them to the proper StorageMachine function. Many of these
+// run at once in different threads. This also makes it such that
+// any StorageMachine implementation must be thread safe.
+func (s *Server) storageDispatchHandler() {
+	for req := range s.storageDispatchCh {
```
maxtaco: is there also a need to listen on the end of the program too, and bail out then?  I've seen it in some of @patrickxb's similar handler loops.

mmaxim: Why?

mmaxim: I'll just add it.

rpc/server/main.go
```diff
@@ -314,6 +320,71 @@ func (s *Server) Serve() error {
...
+
+type syncReq struct {
+	sm     gregor.StateMachine
+	log    rpc.LogOutput
+	arg    gregor1.SyncArg
+	respCh chan<- syncRet
+}
+
+// storageConsumeMessage schedules a Consume request on the dispatch handler
+func (s *Server) storageConsumeMessage(m gregor.Message, retCh chan<- error) {
```
maxtaco: make the `retCh` in this function and then just return `error` when done.

rpc/server/main.go
```diff
@@ -314,6 +320,71 @@ func (s *Server) Serve() error {
...
+	err := s.storageDispatch(req)
+	if err != nil {
+		retCh <- err
+	}
+}
+
+// storageSync schedules a Sync request on the dispatch handler
+func (s *Server) storageSync(sm gregor.StateMachine, log rpc.LogOutput, arg gregor1.SyncArg,
+	retCh chan<- syncRet) {
+	req := syncReq{sm: sm, log: log, arg: arg, respCh: retCh}
```
maxtaco: same thing as above, don't pass in the channel if possible, just make it local to this function.

maxtaco:
This looks good to me.  If you see a chance to simplify some of these methods, you should do it.   Also, please coordinate with @patrickxb about how to handle conflicts in the critiical serving path.

rpc/server/main.go
mmaxim: I took @patrickxb's suggestion and just close the dispatch channel in `Shutdown`

patrickxb:
LGTM

PR #75, opened 2016-05-12
add a String() method to MsgID

oconnor663:
@maxtaco @patrickxb @jacobhaven let me know what you guys think about adding these string methods as we need them? Alternatives include implementing some global interface for anything that has `Bytes()`, or just using `EncodeToString` manually everywhere.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -21,6 +21,7 @@ type UID interface {
 
 type MsgID interface {
 	Bytes() []byte
+	String() string
 }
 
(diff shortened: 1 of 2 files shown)


maxtaco:
May as well do so for the other data types too (MsgID, UID, etc)

oconnor663:
@maxtaco added.

maxtaco:
👍

PR #76, opened 2016-05-12
use a symlink instead of vendoring gregor libs

oconnor663:
More of a request for comments than an actual PR.


patrickxb:
does this work?

maxtaco:
If so we should do it.

oconnor663:
It doesn't break the build, but it does seem to break some tests. Digging...

```
--- FAIL: TestUsage (0.00s)
        Error Trace:    flag_test.go:21
                        flag_test.go:45
        Error:          Expected value not to be nil.
        Messages:       opts came back

--- FAIL: TestRunTLS (0.00s)
        Error Trace:    server_test.go:186
        Error:          Expected nil, but got: "Error parsing session server URI"
        Messages:       no error

--- FAIL: TestRunTCP (0.00s)
        Error Trace:    server_test.go:230
        Error:          Expected nil, but got: "Error parsing session server URI"
        Messages:       no error

FAIL
FAIL    github.com/keybase/gregor/bin/gregord   0.026s
--- FAIL: TestUsage (0.00s)
        Error Trace:    flag_test.go:20
                        flag_test.go:33
        Error:          Expected value not to be nil.
        Messages:       opts came back

FAIL
FAIL    github.com/keybase/gregor/bin/naggerd   0.004s
```

jacobhaven:
@oconnor663 The vendoring appears to work fine. I'm not sure what's causing these errors parsing session server URIs...

oconnor663:
Same thoughts here. I'll see if I can dig into it this weekend.

maxtaco:
I think the symlink broke the Go type system.  According to my printfs, we have an object of type `*rpc.FMPURI` that won't match the type assertion in `bin/config.go`.  It seems fine though, I think the runtime is just fooled by the symlink.

jacobhaven:
@maxtaco @oconnor663 So it seems like this might not be possible given how the Go build system treats symlinks (implicitly flattening them).

oconnor663:
Even if they're flattened, we should have everything going through the flattened, vendored hierarchy. I'm still not sure how we end up with two different types. Haven't had time to look at it yet, but I will after https://keybase.atlassian.net/browse/CORE-2977 is in.

maxtaco:
Another point is that this is likely the first RTTI-dependent branch we hit, but we're likely to see many more as we run the code.

oconnor663:
Latest version uses multiple symlinks, excluding `bin`. Fixes the test failures from before. (I'm not getting clean master runs on my laptop, though, so I have fingers crossed for CI.)

maxtaco:
Same two failing tests

oconnor663:
@maxtaco are you sure? Do you mean the same two tests that were failing before in this PR, or the same tests that are failing in master?

maxtaco:
same two tests are failing in master i believe, but not with 100% probability

maxtaco:
-- FAIL: TestPublish (0.02s)

[I] Authenticate: goodtoken
panic: send on closed channel

these two tests have been broken for about 2 days i think, across all branches

oconnor663:
Yeah, I just did this clean run: https://travis-ci.org/keybase/gregor/builds/131173466

maxtaco:
? Still failing: `AIL: TestPublish (0.03s)
	publish_test.go:73: client broadcasts received: 0, expected 1`

oconnor663:
@maxtaco I see that failure in the clean master run above.

oconnor663:
And in fact, on my local master. I'll bisect it.

oconnor663:
It looks like https://github.com/keybase/gregor/pull/72 broke things?

patrickxb:
@mmaxim said he knew about it here:  https://github.com/keybase/gregor/pull/79

maxtaco:
Just to be sure, we're talking about 2 independent bugs, right?

patrickxb:
Sorry, I misunderstood.

jacobhaven:
So, I think the decision we made was to just vendor at the gregor repo top-level and see if that breaks anything in client (It shouldn't...)

PR #77, opened 2016-05-12
Updates for rsenderd deployment

jacobhaven:
r? @maxtaco


bin/rsenderd/rsender.go
```diff
@@ -0,0 +1,61 @@
...
+
+func newRSender(db *sql.DB, remindServer *rpc.FMPURI, log rpc.LogOutput) *rSender {
+	return &rSender{db: db, sm: storage.NewMySQLEngine(db, of), remindServer: remindServer, log: log}
+}
+
+func (r *rSender) sendReminders() error {
+	remind := r.remind
+	if remind == nil {
+		r.log.Debug("Dialing RemindServer: %s", r.remindServer.String())
+		conn, err := r.remindServer.Dial()
```
maxtaco: this doesn't reconnect.  it should use the same rpc/connection.go mechanism that we're using everywhere else. this will also allow us to put in server<->server authentication.

bin/rsenderd/main.go
```diff
@@ -0,0 +1,35 @@
...
+	db, err := sql.Open("mysql", opts.MysqlDSN)
+	if err != nil {
+		log.Error("%#v", err)
+		os.Exit(2)
+	}
+
+	r := newRSender(db, opts.RemindServer, log)
+	for _ = range time.Tick(opts.RemindDuration) {
+		if err := r.sendReminders(); err != nil {
+			log.Error("%#v", err)
```
maxtaco: Is exiting the right behavior on error?  It's better to log, bumping a stat, and then retry. Exiting will do the same thing but would incur a container restart, which doesn't seem useful here.

maxtaco:
Some comments about failure resilience and retry mechanism

maxtaco:
👍

PR #78, opened 2016-05-16
expand gregor client (CORE-3027)

mmaxim:
@maxtaco The main changes here are to add a `Save` timer (which I think works) and to strip out the incoming interface, and just always make that a parameter to `Sync`. That makes it easier to coordinate in the client.


mmaxim:
Also adds the `replayer` stuff

mmaxim:
@maxtaco this is updated with both my actual changes plus the refactor of the mysql test stuff, let me know if it looks ok.

rpc/client/client.go
```diff
@@ -25,34 +24,20 @@ type Client struct {
...
-		log:      log,
+func NewClient(user gregor.UID, device gregor.DeviceID, sm gregor.StateMachine,
+	storage LocalStorageEngine, log rpc.LogOutput) *Client {
+	c := &Client{
+		user:      user,
+		device:    device,
+		sm:        sm,
+		storage:   storage,
+		log:       log,
+		saveTimer: time.Tick(1 * time.Minute), // How often we save to local storage
```
maxtaco: tunable?

maxtaco:
👍

rpc/client/client.go
jacobhaven: Just `time.Minute`, but yeah I'd agree this should probably be a parameter.

mmaxim:
@maxtaco you want to review this one more time before I merge it?

rpc/client/client.go
```diff
@@ -106,39 +105,69 @@ func (c *Client) syncFromTime(t *time.Time) error {
...
 	}
-	res, err := c.incomingInterface().Sync(ctx, arg)
+
+	// Grab the events from gregord
+	res, err := cli.Sync(ctx, arg)
 	if err != nil {
 		return err
 	}
 
+	// Replay all the messages on the outgoing interface
```
maxtaco: is this comment still relevant?

rpc/client/client.go
```diff
@@ -106,39 +105,69 @@ func (c *Client) syncFromTime(t *time.Time) error {
...
 			c.sm.Clear()
-			err = c.syncFromTime(nil)
+			err = c.syncFromTime(cli, nil)
 		}
 		return err
 	}
 	return nil
 }
+
+func (c *Client) ConsumeMessage(m gregor1.Message) error {
```
maxtaco: remind me, why do we need this again?

mmaxim: Nope...

rpc/client/client.go
mmaxim: Nope...

rpc/client/client.go
```diff
@@ -25,34 +26,32 @@ type Client struct {
...
-		sm:       sm,
-		storage:  storage,
-		incoming: incoming,
-		log:      log,
+func NewClient(user gregor.UID, device gregor.DeviceID, sm gregor.StateMachine,
+	storage LocalStorageEngine, log rpc.LogOutput) *Client {
+
+	// Grab a time interval for flushing state machine to local storage
+	td := time.Duration(time.Minute)
+	etd := os.Getenv("GREGOR_CLI_SAVE_INTERVAL_SECONDS")
```
maxtaco: this doesn't match the way we pass in configuration to the client in other places. Can you make it configurable via argument to `NewClient`?

rpc/client/client.go
mmaxim: To record messages that come from `BroadcastMessage` in the client.

maxtaco: Ok.

rpc/client/client.go
mmaxim: Sure!

test/test_db_sync.go
```diff
@@ -1,51 +1,38 @@
...
+			t.Skip("Error parsing MYSQL_DSN")
 		}
-	}
+		query := udsn.Query()
+		query.Set("parseTime", "true")
+		udsn.RawQuery = query.Encode()
 
-	if err := tx.Commit(); err != nil {
-		return nil, err
+		dsn = udsn.String()
```
maxtaco: i think on second thought it's worth factoring out this code into a common place, so that you can call it from both here and from `bin/config.go`.  So we can reuse the code and not introduce the dependency on bin

maxtaco:
A few more comments

maxtaco:
👍

PR #79, opened 2016-05-17
make MYSQL_DSN errors clearer

oconnor663:
r? @jacobhaven

### DIFF
--- a/bin/dbinit/dbinit.go
+++ b/bin/dbinit/dbinit.go
@@ -11,8 +11,9 @@ import (
 )
 
 var (
+	dsnVar   = "MYSQL_DSN"
 	s3conf   bin.S3Config
-	mysqlDSN = &bin.DSNGetter{S: os.Getenv("MYSQL_DSN"), S3conf: &s3conf}
+	mysqlDSN = &bin.DSNGetter{S: os.Getenv(dsnVar), S3conf: &s3conf}
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


jacobhaven:
I'd sort of prefer to create a `NewDSNGetter(*S3Config)` and move it out of the `bin` package to remove that test dependency.

mmaxim:
My current PR does that @jacobhaven if you want to check out how I moved it out of bin into test. It was necessary since we had to vendor the storage directory into the client.

maxtaco:
@mmaxim you might want to check out the C/I bug above, I think we have a flakey test  / race condition stemming from a PR you pushed last week.

mmaxim:
@maxtaco this one?

=== RUN   TestLevelDBClient
[I] Sync failure: local state hash != server state hash

maxtaco:
No that one is expected.  The close on end channel stack trace at the end of the C/I run. 

@patrickxb do you have any ideas about the lack of broadcast?

patrickxb:
I don't know why the broadcast failed in TestPublish.

mmaxim:
This is just a bug with one of the tests I recently added, basically a `ConsumeMessage` comes in after `Shutdown` gets called, which doesn't seem like something that can happen in normal operation. I'll fix the test.

oconnor663:
@jacobhaven can I go ahead and land this?

jacobhaven:
👍

PR #82, opened 2016-05-19
rename DB tables and session server

maxtaco:
- prefix all DB tables with `gregor_` so that we can, optionally, store this data in an existing database.
- rename session server to auth server

### DIFF
--- a/bin/gregord/config.go
+++ b/bin/gregord/config.go
@@ -15,7 +15,7 @@ import (
 )
 
 type Options struct {
-	SessionServer     *rpc.FMPURI
+	AuthServer        *rpc.FMPURI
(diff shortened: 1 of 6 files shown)


maxtaco:
r? @jacobhaven ? I'm going to need help updating our rollouts

These are two independent renamings, but thought it would make sense to get them done at the same time.

jacobhaven:
:+1: Will require a change in the environment variable used in the AWS deploy script [here](https://github.com/keybase/aws/blob/master/kbfs/lib/kbfs.rb#L263) and some manual schema changes run on the dev and prod DBs.

maxtaco:
Cool thanks for the pointer

PR #85, opened 2016-05-19
Add scare warning to dbinit

jacobhaven:
To prevent accidentally DROPing tables that could contain useful data.

r? @maxtaco


maxtaco:
👍  thanks

PR #86, opened 2016-05-19
Call authd to get super user session token

patrickxb:
This needs changes to keybase/keyabse authd.iced.

r? @maxtaco @mmaxim


rpc/server/main.go
```diff
@@ -358,8 +363,24 @@ func (s *Server) publishProcess() {
...
 	s.log.Debug("creating new aliveGroup")
-	a := <-s.addr
-	s.group = newAliveGroup(s.statusGroup, a.String(), s.authToken, s.publishTimeout, s.clock, s.closeCh, s.log)
+	timeout := 2 * time.Second
+	var a net.Addr
+	select {
+	case a = <-s.addr:
+	case <-time.After(timeout):
+		s.log.Debug("waited %s for address", timeout)
+		panic("timeout waiting for address")
```
maxtaco: panic?

rpc/server/main.go
```diff
@@ -358,8 +363,24 @@ func (s *Server) publishProcess() {
 
 func (s *Server) createAliveGroup() {
 	s.log.Debug("creating new aliveGroup")
-	a := <-s.addr
-	s.group = newAliveGroup(s.statusGroup, a.String(), s.authToken, s.publishTimeout, s.clock, s.closeCh, s.log)
+	timeout := 2 * time.Second
+	var a net.Addr
+	select {
```
maxtaco: can you explain this select with the `panic`? It looks somewhat scary

patrickxb: 1. this shouldn't happen and in the event that it does it would be nice to see the stack trace in the log.
2. the rpc server can't function without an addr or a super token, so if it fails to get either one within a reasonable amount of time, it should quit and the process should be restarted by whatever is managing the gregord process.

maxtaco: What's the lag based on (if there's any?).  RPC to authd?

maxtaco: I worry about cascading failures...

rpc/server/main.go
```diff
@@ -358,8 +363,24 @@ func (s *Server) publishProcess() {
...
+	case a = <-s.addr:
+	case <-time.After(timeout):
+		s.log.Debug("waited %s for address", timeout)
+		panic("timeout waiting for address")
+	}
+
+	var t gregor1.SessionToken
+	select {
+	case t = <-s.superCh:
```
maxtaco: so this one is definitely based on network lag to contact authd, from my read...

rpc/server/authd_handler.go
```diff
@@ -10,25 +10,38 @@ import (
...
+
+	// get a super user token
+	ac := gregor1.AuthInternalClient{Cli: cli}
+	tok, err := ac.CreateGregorSuperUserSessionToken(ctx)
+	if err != nil {
+		a.log.Debug("authd handler: error creating super user session token: %s", err)
+		return err
+	}
+	a.log.Debug("authd handler: created super user session token")
+	a.superCh <- tok
```
maxtaco: btw, this expire after a while.  we might need to renew them too if the system stays up for a while...

rpc/server/main.go
patrickxb: yes.  My other thought was something along the lines of:

    for {
        select {
            case t = <-s.superCh: break
            case <-time.After(1 * time.Second): s.log.Debug("slow getting super token")
        }
    }

This would be the equivalent of a bare `t = <-s.superCh` but would log every second if it was slow/hung.

rpc/server/authd_handler.go
patrickxb: Ok, I didn't know that.  Then this will need to be reworked....

rpc/server/main.go
maxtaco: Yeah, I favor this over crashing....

rpc/server/authd_handler.go
maxtaco: Sorry I neglected to mention it.  Yeah, all session tokens expire after a while....

patrickxb: What's the timeout?

maxtaco: the token itself contains the timeout.  i can change authd to tell you

rpc/server/main.go
patrickxb: The benefit of crashing is it might actually work when restarted...but I will change it to just log for now.

maxtaco: Where does the non-determinism creep in?  This is just the process getting its own address right?  BTW, I thought due to docker, we need to pass that in, since the internal address and the external address might be different.

rpc/server/main.go
maxtaco: BTW, don't we need to pass in the address to resolve the internal v external address problem with docker?  And in that case, won't we know the address at startup?

patrickxb: The address is known.

rpc/server/authd_handler.go
patrickxb: I added a flag for the refresh interval.

patrickxb:
@maxtaco PTAL

rpc/server/authd_handler.go
```diff
@@ -10,25 +10,48 @@ import (
...
+
+	go a.updateSuperToken(conn)
```
maxtaco: This doesn't seem like the right place to launch it, since we'll have multiple refresh loops running if we connect multiple times, won't we?

maxtaco:
this is good, just had one more question

rpc/server/authd_handler.go
patrickxb: Good point.  Needs to be here to get the (new) connection, but I added code to stop the existing loop.

maxtaco:
ok 🚢 after the fix

PR #87, opened 2016-05-20
don't double-hexlify

maxtaco:
- we already have a `String()` method on this guy, so doing that plus `%x` will double-hexlify


patrickxb:
LGTM

PR #88, opened 2016-05-20
child-mode, in which the process runs as long as stdin stays open

maxtaco:
for launching from node and killing cleanly, even on unkillable SIGKILL from forever


patrickxb:
LGTM

PR #89, opened 2016-05-20
Vendor go-sqlite3

akalin-keybase:
This is needed by a test file.

Also stub out travis install step, so that it doesn't
automatically fetch all dependencies, which would
hide vendoring problems.

Also add some tags to the ignore field, so that
govendor doesn't list appengine modules (and
the go 1.7 context module) as missing.


akalin-keybase:
oh, this may be obsolete

maxtaco:
yeah, we licked this problem yesterday!

akalin-keybase:
Okay, redid this PR with some more minor fixes!

maxtaco:
I believe we only use SQLite in test. Does it still need to be vendored?

akalin-keybase:
I think so. In general, I think we want test dependencies to be vendored so that tests are reproducible. client and kbfs already do this, so I think gregor should do it too.

But the question is, why were tests still passing if we weren't vendoring go-sqlite3? And the reason is that client and kbfs stub out the Travis install step, which runs `go get -t ./...`, which installs dependencies, which masks any missing vendored packages. So I did the same here.

What do you think?

maxtaco:
I think we intentionally left SQLite out of it because we were worried it wouldn't build on iOS, since cross compilation of Cgo is nothing we bothered to figured out. But I don't see how vendoring sqlite will affect this.

The only other thing I'll note is that it's a PITA to have SQLite compiling all of the time when we're developing since its slow. Typically I hack around that by running `go install` from the `sqlite` directory.  But I assume that hack is still possible with this change?

akalin-keybase:
Yeah, if we want to build on iOS, then we can just put the test in its own file with a `build !ios` tag or whatever. But assuming we want reproducible tests, we'd still have to vendor all test dependencies across all platforms we care about.

That hack is still possible, but I don't think you need the hack at all; `go install` from gregor should normally cache the result of unvendored `sqlite` in `$GOROOT/pkg`, and similarly, if sqlite is vendored, then it'll cache the result of the vendored `sqlite` in a different directory (a subdir of where the gregor pkg files are).

But if you still want to do `go install` on vendored `sqlite`, you could just do `go install` from the vendor directory.

maxtaco:
I think the hack was needed when doing testing, since `go test` doesn't use the caching logic from `go install`, but apparently there is a flag for that (I forget which)

maxtaco:
Ok thanks for the explanation, let's give it a try
👍

PR #90, opened 2016-05-20
state machine generates a ctime

mmaxim:
@maxtaco @patrickxb @jacobhaven this change adds another return value to `ConsumeMessage` (the ctime) that will come along with `BroadcastMessage` so gregord and the client don't have two different ctime's for the same message.

### DIFF
--- a/rpc/server/main.go
+++ b/rpc/server/main.go
@@ -327,9 +327,11 @@ func (s *Server) startSync(c context.Context, arg gregor1.SyncArg) (gregor1.Sync
 // and broadcasts it to the other clients. Like startSync, this function is called
 // from connection, and is on a different thread than Serve
 func (s *Server) runConsumeMessageMainSequence(c context.Context, m gregor1.Message) error {
-	if err := s.storageConsumeMessage(m); err != nil {
-		return err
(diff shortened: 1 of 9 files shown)


rpc/server/main.go
```diff
@@ -422,12 +429,15 @@ type syncReq struct {
...
-	retCh := make(chan error)
+func (s *Server) storageConsumeMessage(m gregor.Message) consumeMessageRet {
+	retCh := make(chan consumeMessageRet)
 	req := consumeMessageReq{m: m, respCh: retCh}
 	err := s.storageDispatch(req)
 	if err != nil {
-		return err
+		return consumeMessageRet{
+			err:   err,
+			ctime: time.Now(),
```
patrickxb: Can you remove this line in the case of an error?

patrickxb:
LGTM.  one question, but not important.

rpc/server/main.go
mmaxim: Sure!

mmaxim:
@maxtaco can I merge this?

maxtaco:
👍

PR #91, opened 2016-05-23
Read DEBUG environment var to set debug

jacobhaven:
This now matches the environment variable comments now...
r? @patrickxb @maxtaco


maxtaco:
👍

PR #92, opened 2016-05-24
rename server status table to gregor_server_status

maxtaco:
- also don't make it DATETIME(6) since that's not available on 5.5.49
  (which comes with LTS 14.04).

### DIFF
--- a/srvup/storage_mysql.go
+++ b/srvup/storage_mysql.go
@@ -69,9 +69,9 @@ func (s *StorageMysql) UpdateServerStatus(group, hostname string) error {
 	var err error
 	if s.update == nil {
 		if s.clock != nil {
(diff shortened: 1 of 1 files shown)


maxtaco:
Just merging this to fix keybase/keybase C/I

patrickxb:
We use DATETIME(6) in other places in gregor.

maxtaco:
i know i've had to pull it out if running in CI

maxtaco:
ok, let me back that back out, good point.  don't neeed it anyways if we create the tables manually in keybase.sql

patrickxb:
Also, the reason it was `server_status` was so it could be used by any service, not just gregor.

maxtaco:
oh ok, i can back this whole change out and workaround it a different way

patrickxb:
if the name is too vague, i'm fine with another name, just would prefer if it didn't have `gregor` in it.  If we want to use it for `reminderd` for example.

PR #93, opened 2016-05-24
log message with the time of the item

mmaxim:
@maxtaco key move here is having `logMessage` use the ctime from the event if it gets sent in, this is useful for testing.

### DIFF
--- a/rpc/client/client.go
+++ b/rpc/client/client.go
@@ -156,3 +156,7 @@ func (c *Client) StateMachineLatestCTime() *time.Time {
 func (c *Client) StateMachineInBandMessagesSince(t time.Time) ([]gregor.InBandMessage, error) {
 	return c.sm.InBandMessagesSince(c.user, c.device, t)
 }
+
+func (c *Client) StateMachineState(t gregor.TimeOrOffset) (gregor.State, error) {
+	return c.sm.State(c.user, c.device, t)
+}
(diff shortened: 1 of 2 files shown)


rpc/client/client.go
```diff
@@ -156,3 +156,7 @@ func (c *Client) StateMachineLatestCTime() *time.Time {
 func (c *Client) StateMachineInBandMessagesSince(t time.Time) ([]gregor.InBandMessage, error) {
 	return c.sm.InBandMessagesSince(c.user, c.device, t)
 }
+
+func (c *Client) StateMachineState(t gregor.TimeOrOffset) (gregor.State, error) {
+	return c.sm.State(c.user, c.device, t)
+}
```
maxtaco: and what's this?

storage/mem_sm.go
```diff
@@ -252,12 +252,14 @@ func (m *MemEngine) consumeInBandMessage(uid gregor.UID, msg gregor.InBandMessag
...
 	default:
 	}
-	user.logMessage(now, msg, i)
 
 	retTime := now
 	if i != nil {
 		retTime = i.ctime
 	}
+
+	user.logMessage(retTime, msg, i)
```
maxtaco: got it 👍

maxtaco:
had a brief q

rpc/client/client.go
mmaxim: It was convenient to have for my tests in my keybase client PR

maxtaco: Ok thumbs up!

PR #94, opened 2016-05-26
return the messages from Sync to the caller CORE-3113

mmaxim:
r? @maxtaco

This is a simple change to just get the list of messages that we consume in a sync operation back to the caller. It is useful for unit tests against this system on the client.


maxtaco:
👍

PR #97, opened 2016-05-31
use IDs instead of hostnames CORE-3105

mmaxim:
@maxtaco @patrickxb this patch introduces the use of IDs to uniquely identify gregord nodes, instead of hostname/port. The IDs are just a random base64 encoded string that is stored in the DB along with the host/port pair. 

In addition, the patch stops threading the context of the caller into gregord through to both broadcast and publish. Those operations really are not part of the client's request, and it seems more appropriate to attach them to the background context. Otherwise, you get weird results if the client hangs up early (i.e. you don't publish the consume, which doesn't seem right).

This is a big patch, so let me know what you think!

### DIFF
--- a/srvup/srvup.go
+++ b/srvup/srvup.go
@@ -5,12 +5,13 @@ package srvup
 
 import (
 	"crypto/rand"
-	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/jonboulle/clockwork"
+	rpc "github.com/keybase/go-framed-msgpack-rpc"
 )
 
 // NodeId is a random id of a node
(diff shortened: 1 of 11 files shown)


srvup/srvup.go
```diff
@@ -4,39 +4,63 @@
...
 
 	"github.com/jonboulle/clockwork"
 )
 
+// NodeId is a random id of a node
+type NodeId string
+
+// NodeDesc identifies a node in the group
+type NodeDesc struct {
+	Hostname string
```
mmaxim: Change to hostname/pair name or URI

srvup/srvup.go
```diff
@@ -4,39 +4,63 @@
...
 	wg                 sync.WaitGroup
 }
 
 // New creates a new Status for a server.
 func New(group string, heartbeatInterval, aliveThreshold time.Duration, s Storage) *Status {
+
+	// Generate a random ID for ourselves
+	rawid := make([]byte, 8)
+	rand.Read(rawid)
+	id := base64.StdEncoding.EncodeToString(rawid)
```
mmaxim: Change to hex encoding

mmaxim: Also output some debug about our ID

srvup/storage_mysql.go
```diff
@@ -180,12 +180,13 @@ func (s *StorageMysql) resetSchema() error {
 
 var schema = []string{
 	`CREATE TABLE IF NOT EXISTS server_status (
```
mmaxim: Change the ID and name here too

maxtaco:
👍 after naming fixes

PR #98, opened 2016-05-31
validate messages in ConsumeMessage CORE-3119

mmaxim:
@maxtaco @oconnor663 This patch addresses CORE-3119 to make it harder for gregord to crash on bad input to its incoming RPCs. 

@oconnor663 you probably want to double check this won't break anything with the tracker popup window stuff, just to make sure it is passing all the required fields.

### DIFF
--- a/rpc/server/connection.go
+++ b/rpc/server/connection.go
@@ -167,6 +167,9 @@ func validateConsumeMessage(m gregor1.Message) error {
 	}
 	ibm := m.ToInBandMessage()
 	obm := m.ToOutOfBandMessage()
+	if ibm == nil && obm == nil {
+		return errfunc("unknown message type")
+	}
 	if ibm != nil {
 		upd := ibm.ToStateUpdateMessage()
 		if upd != nil {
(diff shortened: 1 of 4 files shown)


rpc/server/connection.go
```diff
@@ -160,8 +160,46 @@ func (c *connection) Sync(ctx context.Context, arg gregor1.SyncArg) (gregor1.Syn
 	return c.parent.startSync(ctx, arg)
 }
 
+func validateConsumeMessage(m gregor1.Message) error {
+	err := errors.New("invalid consume call, missing required fields")
```
oconnor663: It would be nice if we could either put the field name in the error itself (maybe with a little lambda here that takes the name and formats the error) or log the field name right before we return the error.

rpc/server/connection.go
```diff
@@ -160,8 +160,46 @@ func (c *connection) Sync(ctx context.Context, arg gregor1.SyncArg) (gregor1.Syn
...
+		upd := ibm.ToStateUpdateMessage()
+		if upd != nil {
+			if upd.Metadata() == nil || upd.Metadata().MsgID() == nil || upd.Metadata().UID() == nil {
+				return err
+			}
+			if upd.Creation() != nil {
+				if upd.Creation().Category() == nil || upd.Creation().Body() == nil {
+					return err
+				}
+			} else if upd.Dismissal() != nil {
```
oconnor663: Should we validate that a message doesn't contain _both_ and creation _and_ a dismissal? Alternatively, if that's legal, then we might need to check the validity of both sides, rather than skipping the dismissal side after we see that the creation exists.

rpc/server/connection.go
```diff
@@ -160,8 +160,46 @@ func (c *connection) Sync(ctx context.Context, arg gregor1.SyncArg) (gregor1.Syn
...
+				if upd.Dismissal().MsgIDsToDismiss() == nil || upd.Dismissal().RangesToDismiss() == nil {
```
oconnor663: Similar to above, is it legal to dismiss both by ID and by range at the same time?

rpc/server/connection.go
```diff
@@ -160,8 +160,46 @@ func (c *connection) Sync(ctx context.Context, arg gregor1.SyncArg) (gregor1.Syn
...
+					return err
+				}
+			} else {
+				return err
+			}
+		}
+	} else if obm != nil {
```
oconnor663: Similar to above, how do we feel about being both an ibm and an obm at the same time? Probably that's not good?

oconnor663:
I think we're good. All the real dismissals generated by the client go through here: https://github.com/keybase/client/blob/522ebfc142e8b28da8f8ca2a2892c5ff6c557f63/go/service/gregor.go#L671-L683

The server also generates messages, here: https://github.com/keybase/keybase/blob/master/lib/mod/gregor.iced#L87-L91 I think that fills out similar fields.

What does `gregord` do if these fields are missing? I just realized I have no idea :|

mmaxim:
@oconnor663 Thanks! The point of the function was mostly just to ensure that the message was well-formed enough that it wouldn't cause gregord to crash later on. The cases where a message might have both of something (in band and out of band, or create and dismiss) was something I thought could be handled deeper. Although I suppose that if gregord does get a message that is both in band and out of band it won't validate the out of band part.

mmaxim:
@maxtaco @oconnor663 what do you think here, good to go?

maxtaco:
👍

oconnor663:
:+1:

PR #99, opened 2016-06-01
Moved Dockerfile and added dbinit

jzila:


jzila:
@mmaxim ready for review

mmaxim:
👍 Looks good! I wonder if at some point we should have the `server_status` table get created by `dbinit` instead of `gregord` itself, but that can be a separate change I think (and shouldn't affect this).

PR #102, opened 2016-06-02
improve gregor logging CORE-3167

mmaxim:
@maxtaco ?

### DIFF
--- a/bin/log.go
+++ b/bin/log.go
@@ -26,6 +26,9 @@ type StandardLogger struct {
 
 var _ rpc.LogOutput = (*StandardLogger)(nil)
 
+// appendID will add in the node ID of the current gregor to all log output.
+// Note that there can be tough if g.myID has a % in it, but that should
+// never happen since all IDs are hex encoded.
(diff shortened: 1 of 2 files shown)


mmaxim:
log line now looks like this:

13:10:10.849 ▶ DEBU gregord log.go:261 063 [b4ead8d43a0a76e9] {tcp://127.0.0.1:51598} reply(11): method=gregor.1.incoming.ping; err=null; res="pong";

maxtaco:
i guess it's kosher but it's sneaky!

bin/log.go
```diff
@@ -20,24 +21,36 @@ import (
...
 	inner  *logging.Logger
 	module string
+	myID   *srvup.NodeId
 }
 
 var _ rpc.LogOutput = (*StandardLogger)(nil)
 
+func (g *StandardLogger) appendID(s string) string {
+	if g.myID != nil {
+		return fmt.Sprintf("[%s] %s", string(*g.myID), s)
```
maxtaco: It's fine if you add some comments.  it feels dangerous but it works out

PR #104, opened 2016-06-03
pubsub uses TLS now

mmaxim:
Refactor things so that pubsub connects using a framed-msgpack URI instead of a host/port address. This allows for each gregor to write into `server_status` the correct protocol (TLS or not) it used when setting up the listen loop.


maxtaco:
👍

PR #105, opened 2016-06-03
add check to see if gregor tables exist before recreating CORE-3169

mmaxim:
@maxtaco @jzila this is pretty simple.

### DIFF
--- a/bin/dbinit/dbinit.go
+++ b/bin/dbinit/dbinit.go
@@ -14,14 +14,16 @@ import (
 )
 
 var (
-	skipCheck bool
-	dsnVar    = "MYSQL_DSN"
-	s3conf    bin.S3Config
-	mysqlDSN  = &bin.DSNGetter{S: os.Getenv(dsnVar), S3conf: &s3conf}
+	skipSanityCheck bool
+	skipExistsCheck bool
+	dsnVar          = "MYSQL_DSN"
+	s3conf          bin.S3Config
+	mysqlDSN        = &bin.DSNGetter{S: os.Getenv(dsnVar), S3conf: &s3conf}
 )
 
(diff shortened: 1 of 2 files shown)


maxtaco:
👍

bin/dbinit/dbinit.go
```diff
@@ -66,6 +68,17 @@ func main() {
...
 	}
 
+	// Check if the gregor tables already exist, unless we are just charging forward
+	if !skipExistsCheck {
+		exists, err := schema.DBExists("mysql", dsn)
+		if err != nil {
+			log.Fatal(err)
+		}
+		if exists {
+			log.Fatal("database already exists, exiting...")
```
jzila: shouldn't this behave idempotently instead? I.e. if it exists, do nothing

mmaxim: As opposed to doing `log.Fatal`?

jzila: correct, since log.Fatal will exit

mmaxim: Yeah I thought it would just be easier to hook up if the check happened in dbinit.go, the idempotency is just in the dbinit program, not the schema package. `log.Fatal` gets us out before calling `CreateDB`

jzila: Right but then `dbinit` will return a nonzero result, causing any scripts calling it to fail out.

jzila: So maybe you can call `os.Exit(0)`

mmaxim: Ok I'll change it to that, thanks!

jzila:
lgtm

PR #106, opened 2016-06-04
kill rsenderd and expose functionality via RPC from gregor

maxtaco:
- this will simplify deployment
- lots of various cleanups in Gregor.


mmaxim:
The change looks good to me, but I was just thinking I wonder if it makes sense to just pull that `ReminderProtocol` off into another server (reminderd). It wouldn't be like rsenderd that actually sent messages, but would do basically everything in this commit in a separate process. The thinking here is that this functionality serves no purpose exposed to the world (as evidenced by the super user checks), so why not just drop it on another service that only listens inside AWS?

Anyway, just a thought, but otherwise this looks awesome!

PR #107, opened 2016-06-04
fix up the TLS situation in sibconnections

maxtaco:
- plumb a tls Config from the command line all the way through
- make changes to RPC library directly in vendor for a test-of-concept. If we like this we can go back and
  revendor these changes
- as much as possible, use the built-in Go mechanisms for doing what we were doing by hand
- ~~tests mainly pass, but TestRunTPC is mysteriously horked~~
  - ~~i can debug it, but it might be easier to get an assist from @mikem or @patrickxb to
    see if there's something obvious i've done wrong. the sympton is that we churn
    indefinitely while waiting for a super token~~
- tests all pass!


maxtaco:
@mmaxim can you review this? If we adopt this approach, we need to change the AWS rollout scripts.  But we should already have the root CA in an S3 bucket, so it should be easier in the long run.

rpc/server/sibconn.go
```diff
@@ -64,17 +65,10 @@ func (s *sibConn) Shutdown() {
...
-	if env == "" {
-		env = "staging"
-	}
-
 	// Connect to our peer using proper protocol
 	if s.uri.UseTLS() {
 		s.debug("connecting to gregord (tls)")
-		s.conn = rpc.NewTLSConnectionWithServerName(s.uri.HostPort, gregor.TLSHostNames[env],
-			[]byte(gregor.BundledCAs[env]), keybase1.ErrorUnwrapper{}, s, true,
+		s.conn = rpc.NewTLSConnectionWithTLSConfig(s.uri.HostPort, s.tlsConfig, keybase1.ErrorUnwrapper{}, s, true,
```
mmaxim: Is there a PR for this on the msgpack repo?

mmaxim:
👍 I can update the deployment stuff tomorrow if you want to get this live.

rpc/server/sibconn.go
maxtaco: no, i was going to see if you thought this made sense, and then make a PR over there...

mmaxim: Makes sense, I think this is definitely better, sorry for not doing this in the first place!

maxtaco:
There's a PR out in keybase/go-framed-msgpack-rpc#95 for the vendored configuration that I hacked in

PR #108, opened 2016-06-08
dont block sending super tokens, and make the alive group earlier

mmaxim:
@maxtaco @patrickxb this fixes the gregor deadlock problem.


patrickxb:
LGTM, but travis failed?

maxtaco:
this isn't going to work

PR #109, opened 2016-06-08
fix gregord authd deadlock CORE-3186

mmaxim:
@maxtaco @patrickxb a better attempt, this one removes the usage of superCh being passed throughout the modules, and instead keeps all the synchronization local to authd_handler.go

### DIFF
--- a/rpc/server/authd_handler.go
+++ b/rpc/server/authd_handler.go
@@ -13,7 +13,7 @@ import (
 type authdHandler struct {
 	authserver      gregor1.AuthUpdateInterface
 	log             rpc.LogOutput
-	superCh         chan gregor1.SessionToken
+	superCh         chan struct{}
 	refreshInterval time.Duration
 	doneCh          chan struct{}
 	superToken      gregor1.SessionToken
@@ -32,7 +32,7 @@ func NewAuthdHandler(authserver gregor1.AuthUpdateInterface, log rpc.LogOutput,
 	return &authdHandler{
 		authserver:      authserver,
 		log:             log,
-		superCh:         make(chan gregor1.SessionToken),
+		superCh:         make(chan struct{}),
 		refreshInterval: refreshInterval,
 		doneCh:          make(chan struct{}),
 	}
@@ -46,12 +46,12 @@ func (a *authdHandler) GetSuperToken() gregor1.SessionToken {
 	// If we don't have a super token yet, wait for one
 	if blank {
 		a.log.Debug("authd handler: GetSuperToken(): token blank: waiting...")
-		return <-a.superCh
-	} else {
-		a.Lock()
-		defer a.Unlock()
-		return a.superToken
+		<-a.superCh
 	}
+
+	a.Lock()
+	defer a.Unlock()
+	return a.superToken
 }
 
 func (a *authdHandler) setSuperToken(cli gregor1.AuthInternalClient) error {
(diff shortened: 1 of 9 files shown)


rpc/server/authd_handler.go
```diff
@@ -24,31 +27,72 @@ var _ gregor1.AuthUpdateInterface = (*authdHandler)(nil)
...
 
+func (a *authdHandler) GetSuperToken() gregor1.SessionToken {
+	a.Lock()
+	blank := (a.superToken == "")
+	a.Unlock()
+
+	// If we don't have a super token yet, wait for one
+	if blank {
+		a.log.Debug("authd handler: GetSuperToken(): token blank: waiting...")
+		return <-a.superCh
```
maxtaco: What if multiple people wind up waiting here? Only the first one will be woken  up.

mmaxim: Ugh yeah, ok I can fix that.

maxtaco: this seems like a common pattern.  a bunch of people waiting for something to be set. @patrickxb is there a canonical go way to handle this?

mmaxim: I was just going to close the channel

maxtaco: that can work, but then you need to grab the token in a separate code block, which is fine.

maxtaco: also, where is this thing updated?

mmaxim: In setSuperToken

patrickxb: let me read this in detail, but based on your comment, check out sync.Cond. 

https://golang.org/pkg/sync/#Cond

maxtaco: duh, yeah ok

maxtaco: mike, maybe the right idea is:

```
func (a *authdHandler) GetSuperToken() gregor1.SessionToken {
    <- a.superCh
    lock()
   defer unlock()
    return a.superToken
}
```

maxtaco: and then in the other go routine, `close(a.superCh)` as you suggested.

maxtaco: Saw that, but maybe you don't have to bother to check for blank, you can always wait.

mmaxim: Got it, I wasn't sure what happened on a read from a closed channel, but it looks like it is ok.

mmaxim: Although it does allow me to output more debug info

rpc/server/authd_handler.go
```diff
@@ -24,31 +27,68 @@ var _ gregor1.AuthUpdateInterface = (*authdHandler)(nil)
...
+	return a.superToken
+}
+
+func (a *authdHandler) setSuperToken(cli gregor1.AuthInternalClient) error {
+
+	tok, err := cli.CreateGregorSuperUserSessionToken(context.Background())
+	if err != nil {
+		a.log.Debug("setSuperToken: error creating super user session token: %s", err)
+		return err
+	} else {
```
maxtaco: Can you check for non-blank token?

maxtaco:
👍

patrickxb:
superCh might be a confusing name later on now that it is doing something different than actually containing the super token.

mmaxim:
How about superReceivedCh?

patrickxb:
sounds good

PR #111, opened 2016-06-09
add gregor stats CORE-3075

mmaxim:
@maxtaco @patrickxb 

Some code to get stats into gregor. Let me know if this looks ok, and if there are any stats missing. Thanks!

### DIFF
--- a/rpc/server/connection.go
+++ b/rpc/server/connection.go
@@ -376,8 +376,9 @@ func (c instrumentedConnection) instrument(name string) func() {
 	return func() {
 		dur := time.Since(now)
 		durms := int(dur.Nanoseconds() / 1000000)
-		if durms > 2000 {
+		if durms > 10000 {
 			c.conn.stats.Count("slow response - " + name)
+			c.conn.log.Error("slow response time: rpc: %s time: %dms", name, durms)
 		}
 		c.conn.stats.ValueInt("speed - "+name, durms)
 	}
--- a/rpc/server/connection.go
+++ b/rpc/server/connection.go
@@ -11,6 +11,7 @@ import (
 	keybase1 "github.com/keybase/client/go/protocol"
 	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"github.com/keybase/gregor/protocol/gregor1"
+	"github.com/keybase/gregor/stats"
 	"golang.org/x/net/context"
 )
 
@@ -67,7 +68,13 @@ type connection struct {
 	// startAuthentication() finishes successfully.
 	serverDoneCh <-chan struct{}
 
-	log rpc.LogOutput
+	log   rpc.LogOutput
+	stats stats.Registry
+}
+
+// instrumentedConnection times all of the RPC calls served for stats purposes
+type instrumentedConnection struct {
+	conn *connection
 }
 
 func newConnection(c net.Conn, parent *Server) (*connection, error) {
@@ -79,6 +86,7 @@ func newConnection(c net.Conn, parent *Server) (*connection, error) {
 		parent: parent,
 		authCh: make(chan error, 1),
 		log:    parent.log,
+		stats:  parent.stats.SetPrefix("connection"),
 	}
 
 	if err := conn.startRPCServer(); err != nil {
@@ -130,6 +138,7 @@ func (c *connection) checkMessageAuth(ctx context.Context, m gregor1.Message) er
 
 func (c *connection) AuthenticateSessionToken(ctx context.Context, tok gregor1.SessionToken) (gregor1.AuthResult, error) {
 
+	c.stats.Count("AuthenticateSessionToken")
 	c.log.Debug("Authenticate: %+v", tok)
 	if tok == "" {
 		var UnauthenticatedSessionError = keybase1.Status{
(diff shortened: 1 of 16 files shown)


mmaxim:
quick update: I added batching support to the StatHat go library (in the vendored version in this PR) and integrated it into gregor. @patrickxb you can take a look, and if it makes sense I can create a PR on the main stathat/go repo.

rpc/server/main.go
```diff
@@ -189,6 +199,24 @@ func (s *Server) uidKey(u gregor.UID) (string, error) {
 	return hex.EncodeToString(tuid), nil
 }
 
+func (s *Server) updateServerValueStats() {
+	s.stats.ValueInt("total user servers", len(s.users))
+	totalConns := 0
+	for _, usrv := range s.users {
+		rchan := make(chan statsRes)
+		usrv.statsCh <- rchan
```
maxtaco: what happens if we block here, we hork sstats?

rpc/server/main.go
```diff
@@ -189,6 +199,24 @@ func (s *Server) uidKey(u gregor.UID) (string, error) {
...
+		totalConns += (<-rchan).totalConns
```
maxtaco: so flexibility here for future per-user-server stats?

rpc/server/main.go
mmaxim: Yeah stats get horked, but nothing else. It shouldn't block either, I think that flow is pretty safe. Also, if it blocks here, it means a bunch of other stuff is hosed.

rpc/server/main.go
mmaxim: That was the thinking, in case there are any other stats from user_server that might be interesting, but that we have to access through a channel.

maxtaco:
didn't comb through it with too much detail, but LGTM!

rpc/server/connection.go
```diff
@@ -349,5 +371,70 @@ func (c *connection) close() {
 	c.c.Close()
 }
 
+func (c instrumentedConnection) instrument(name string) func() {
+	now := time.Now()
+	return func() {
+		dur := time.Since(now)
+		durms := int(dur.Nanoseconds() / 1000000)
```
patrickxb: dur / time.Millisecond

rpc/server/connection.go
```diff
@@ -349,5 +371,70 @@ func (c *connection) close() {
...
+			c.conn.stats.Count("slow response - " + name)
+			c.conn.log.Error("slow response time: rpc: %s time: %dms", name, durms)
+		}
+		c.conn.stats.ValueInt("speed - "+name, durms)
+	}
+}
+
+func (c instrumentedConnection) AuthenticateSessionToken(ctx context.Context, arg gregor1.SessionToken) (gregor1.AuthResult, error) {
+	f := c.instrument("AuthenticateSessionToken")
+	defer f()
```
patrickxb: defer c.instrument("AuthenticateSessionToken")()

rpc/server/main.go
```diff
@@ -189,6 +199,24 @@ func (s *Server) uidKey(u gregor.UID) (string, error) {
...
+	}
+	s.stats.ValueInt("total connections", totalConns)
+}
+
+func (s *Server) updateServerValueStatsLoop() {
+	for {
```
patrickxb: might want a way for this loop to end...in practice, when we have these dangling goroutines the test panic output can get messy.

rpc/server/session_cacher.go
```diff
@@ -63,6 +70,17 @@ func (sc *SessionCacher) GetSuperToken() gregor1.SessionToken {
 	return sc.authdHandler.GetSuperToken()
 }
 
+func (sc *SessionCacher) updateStats() {
+	sc.stats.ValueInt("cache size", sc.Size())
+}
+
+func (sc *SessionCacher) updateStatsLoop() {
+	for {
```
patrickxb: might want a way to end this loop too.

vendor/github.com/stathat/go/stathat.go
```diff
@@ -0,0 +1,573 @@
...
+func (br *BatchReporter) PostEZCountOne(statName, ezkey string) error {
+	return br.PostEZCount(statName, ezkey, 1)
+}
+
+func (br *BatchReporter) PostEZValue(statName, ezkey string, value float64) error {
+	br.Lock()
+	defer br.Unlock()
+
+	// Update value cache
+	br.getEZCache(ezkey).valueStats[statName] = value
```
patrickxb: This should average all the values reported over the interval, not replace.

patrickxb:
I had a few minor comments on the gregor side.  

Happy to accept the batch change as a stathat PR with the value average change + wait until finished.  Thanks!

Overall, looks great!

vendor/github.com/stathat/go/stathat.go
```diff
@@ -0,0 +1,573 @@
...
+
+	return nil
+}
+
+func (br *BatchReporter) batchPost() {
+	br.Lock()
```
patrickxb: I'm a little worried about this lock as it stops the world...I know the calls to PostEZ* are async, but...

Would something like:

    var cacheCopy
    lock
    cacheCopy = cache
    cache = make(map cache)
    unlock
    post stats using cacheCopy

work?

rpc/server/connection.go
mmaxim: Yep!

rpc/server/connection.go
mmaxim: That works? Wouldn't this just call instrument at the end of the function?

patrickxb: note the second set of parentheses...

mmaxim: Ok gotcha, I was just confused how `defer` works.

mmaxim:
Thanks @patrickxb! I'm going to make a separate PR on the stathat repo with the changes you suggest.

PR #113, opened 2016-06-14
move avdl path so they can be vendored...

maxtaco:
...and used to compile JS in the keybase client


mmaxim:
Seems fine to me, but just so I'm clear, the problem was that govendor would not vendor the `avdl` directory because there were no Go source files in there? So by putting them in the gregor1 directory they just come along for the ride?

maxtaco:
Yeah, that's exactly right.  Alternative, we could have a put a `0doc.go` style dummy file the avdl directory, but this is slightly cleaner though still ugly!

mmaxim:
Yeah this is better than that alternative for sure, LGTM!

patrickxb:
LGTM

maxtaco:
This is still in play so I won't merge until we know.

PR #117, opened 2016-06-21
convert session cacher into using locks

mmaxim:
@maxtaco not a priority, but this converts `SessionCacher` to using locks instead of the channel loop, it's just much clearer this way.

### DIFF
--- a/rpc/server/session_cacher.go
+++ b/rpc/server/session_cacher.go
@@ -1,6 +1,7 @@
 package rpc
 
 import (
+	"sync"
 	"time"
 
 	"github.com/jonboulle/clockwork"
@@ -11,11 +12,6 @@ import (
 	"golang.org/x/net/context"
 )
 
-type authResp struct {
-	res    gregor1.AuthResult
-	expiry time.Time
-}
-
 type expirySt struct {
 	sid    gregor1.SessionID
 	expiry time.Time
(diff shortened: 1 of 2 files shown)


maxtaco:
I'm on board, but @patrickxb can you give this a second set of eyes?

Also, does it pass `go vet`?

rpc/server/session_cacher.go
```diff
@@ -138,78 +125,56 @@ func (sc *SessionCacher) clearExpiryQueue() {
...
 // AuthenticateSessionToken authenticates a given session token, first against
 // the cache and the, if that fails, against the parent AuthInterface.
 func (sc *SessionCacher) AuthenticateSessionToken(ctx context.Context, tok gregor1.SessionToken) (res gregor1.AuthResult, err error) {
 
-	respCh := make(chan *gregor1.AuthResult)
-
 	// Fire off request for a cached session
 	sc.stats.Count("AuthenticateSessionToken")
-	sc.reqCh <- readTokReq{tok, respCh}
+	sc.RLock()
```
maxtaco: you can wind up with multiple guys working on `tok` at the same time here.  I guess that's not the end of the world, and the old code had this problem too?

mmaxim:
`go vet` seems cool with it.

rpc/server/session_cacher.go
```diff
@@ -112,21 +98,22 @@ func (sc *SessionCacher) deleteSID(sid gregor1.SessionID) {
...
 }
 
 // clearExpiryQueue removes all currently expired sessions.
 func (sc *SessionCacher) clearExpiryQueue() {
+	sc.Lock()
+	defer sc.Unlock()
+
 	now := sc.cl.Now()
 	for _, e := range sc.expiryQueue {
 		if now.Before(e.expiry) {
```
patrickxb: I know this isn't part of this PR, but `sc.deleteSID(e.sid)` seems clearer to me...

rpc/server/session_cacher.go
```diff
@@ -79,30 +76,19 @@ func (sc *SessionCacher) updateStats() {
...
 	sc.sessionIDs[res.Sid] = tok
 }
 
-type deleteSIDReq struct {
-	sid gregor1.SessionID
-}
-
 func (sc *SessionCacher) deleteSID(sid gregor1.SessionID) {
 	if tok, ok := sc.sessionIDs[sid]; ok {
 		if _, ok := sc.sessions[tok]; ok {
```
patrickxb: again, not in this PR, but just `delete(sc.sessions, tok)` is fine.  no need to check if it is already there.

patrickxb:
LGTM!  Much easier to follow.

PR #119, opened 2016-06-22
fix broken dismissal error checking

maxtaco:
it was being too strict before

### DIFF
--- a/rpc/server/connection.go
+++ b/rpc/server/connection.go
@@ -208,13 +208,8 @@ func validateConsumeMessage(m gregor1.Message) error {
 						return errfunc("missing body")
 					}
 				}
-				if upd.Dismissal() != nil {
-					if upd.Dismissal().MsgIDsToDismiss() == nil {
-						return errfunc("missing msg IDs to dismiss")
-					}
-					if upd.Dismissal().RangesToDismiss() == nil {
-						return errfunc("missing ranges to dismiss")
-					}
+				if upd.Dismissal() != nil && upd.Dismissal().MsgIDsToDismiss() == nil && upd.Dismissal().RangesToDismiss() == nil {
+					return errfunc("missing ranges or IDs to dismiss")
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


mmaxim:
Yeah I saw this earlier today, although my fix was to change the client to pass a blank ranges to dismiss (I didn't commit it yet). The reason I did it this was is that the MySQL storage engine just references ranges to dismiss without checking it:

https://github.com/keybase/gregor/blob/master/storage/sql_sm.go#L300-L302

The check was supposed to protect it, but maybe it just needs to get changed.

maxtaco:
I tried empty ranges but they didn't work --- they get converted to `nil` in extras.go.

maxtaco:
BTW, this bug is blocking https://github.com/keybase/keybase/pull/753, preventing a clean C/I run there.

mmaxim:
Ok this is fine then if it is a blocker. Although be aware this makes it such that you could DoS Gregor by just sending these blank values through and crashing it, so something should change at some point (either extras.go or sql_sm.go). 

Why not just change that line in extras.go to be `ret := make([]gregor.MsgRange, 0)`, it seems like a bug to return `nil` for a blank list?

maxtaco:
Weirdly it wasn't crashing for me after I made my fix and sent `nil` or empty ranges...

maxtaco:
I think it's fine to do `range` over a `nil` value. That doesn't seem to cause a crash in my tests. Is that what you were concerned about?

mmaxim:
Yeah I guess the only place that gets used is in `range`, so I guess it is ok.

mmaxim:
I'm still getting used to the fact that in Go you can use these list operations on `nil` and it's all good. Sorry Max! 

👍

maxtaco:
Me too, it blew my mind earlier!

No worries thanks for your late night review!!!!

maxtaco:
Well no rush anyways, seems like jenkins is on the fritz still

PR #121, opened 2016-06-22
refactor UID server sychronization CORE-3237

mmaxim:
@maxtaco @patrickxb 

The major change here is to use locks to synchronize access within both `rpc.server.Server` and `rpc.server.perUIDServer`. The net result is the amount of channel communication between them for managing new connections and such has been greatly reduced. I think this makes the code easier to follow, and less bug prone (although both files now have comments indicating which functions need locks to be called...)

Anyway, this passes `go test -race` and seems to work, but the question is do we like it more. Let me know!

### DIFF
--- a/rpc/server/main.go
+++ b/rpc/server/main.go
@@ -228,16 +228,18 @@ func (s *Server) waitOnUIDServer(usrv *perUIDServer) {
 		select {
 		case <-usrv.ShouldShutdown():
 			s.Lock()
-			if usrv.ConfirmShutdown() {
+			confirmed := usrv.ConfirmShutdown()
+			if confirmed {
 				delete(s.users, s.uidKey(usrv.uid))
-				s.Unlock()
+			}
+			s.Unlock()
+			if confirmed {
 				usrv.Shutdown()
 				if s.events != nil {
 					s.events.UIDServerDestroyed(usrv.uid)
 				}
 				return
 			}
-			s.Unlock()
 		case <-s.shutdownCh:
 			return
 		}
--- a/rpc/server/main.go
+++ b/rpc/server/main.go
@@ -195,7 +195,7 @@ func (s *Server) updateServerValueStatsLoop() {
 		select {
 		case <-s.shutdownCh:
 			return
-		case <-time.After(time.Second):
+		case <-s.clock.After(time.Second):
 			s.updateServerValueStats()
 		}
 	}
--- a/rpc/server/main.go
+++ b/rpc/server/main.go
@@ -68,6 +68,8 @@ type Authenticator interface {
 // Server is an RPC server that implements gregor.NetworkInterfaceOutgoing
 // and gregor.NetworkInterface.
 type Server struct {
+	sync.RWMutex
+
 	// At first we only allow one state machine, but there might be a need to
 	// chain them together.
 	storage gregor.StateMachine
@@ -112,9 +114,6 @@ type Server struct {
 
 	// TLS Config for reaching out to other similar peers
 	tlsConfig *tls.Config
-
-	// UID server lock
-	sync.RWMutex
 }
 
 type ServerOpts struct {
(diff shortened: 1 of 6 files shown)


maxtaco:
@patrickxb you and I both should look at this one

rpc/server/user_server.go
```diff
@@ -74,75 +73,168 @@ func newPerUIDServer(uid gregor1.UID, parentConfirmCh chan confirmUIDShutdownArg
...
 	var wg sync.WaitGroup
 
+	conns := s.connections()
 	s.stats.Count("broadcast")
-	s.stats.ValueInt("broadcast - conns", len(s.conns))
+	s.stats.ValueInt("broadcast - conns", len(conns))
 
-	for id, conn := range s.conns {
-		s.log.Info("uid %s broadcast to %d", s.uid, id)
+	for conn, _ := range conns {
```
patrickxb: for conn := range conns

rpc/server/user_server.go
```diff
@@ -74,75 +73,168 @@ func newPerUIDServer(uid gregor1.UID, parentConfirmCh chan confirmUIDShutdownArg
...
-		s.stats.Count("destroyed")
-	}()
-	for {
+		s.log.Info("uid server %s: connection closed", s.uid)
+		s.removeConnection(conn)
+	}
+}
+
+// connections makes a copy of all the current connections and returns it
+func (s *perUIDServer) connections() map[*connection]bool {
```
patrickxb: why do you need to copy the map?  in case it is modified while you are using it?  can you just return []*connection instead?  it doesn't look like the bool value is ever used by callers of this.

mmaxim: So I didn't need to lock `broadcast` from beginning to end.

patrickxb: Ok, then do you need to return a map?  Or would a slice suffice?

patrickxb:
LGTM

@mmaxim do you think the tests handle this:

> If this is the last connection for the user, the per-user goroutine sends a message to the main goroutine, who cleans out the per-user goroutine and state, and also unsubscribes from pubsub. The main thread will serialize EOFs and new connections so that we don't race when a new connection happens around the same time the last connection for a user dies. Note this is the trickiest case, we should make sure we have a solution for it first!

(from max's doc https://github.com/keybase/gregor/blob/master/doc/code.md)

I remember a lot of issues occurred with the original implementation, so just want to make sure.

rpc/server/main.go
```diff
@@ -126,6 +112,9 @@ type Server struct {
 
 	// TLS Config for reaching out to other similar peers
 	tlsConfig *tls.Config
+
+	// UID server lock
+	sync.RWMutex
```
maxtaco: the go convention is to move this struct member closer to the thing it's protecting.

rpc/server/main.go
```diff
@@ -498,39 +430,56 @@ func (s *Server) publish(m gregor1.Message) error {
...
+		s.log.Error("broadcastConsumeMessage: failed to get UID server: %s", err)
+		return
+	}
+
+	// Nothing to do...
+	if srv == nil {
+		// even though nothing to do, create an event if
+		// an event handler in place:
+		if s.events != nil {
+			s.log.Info("broadcastConsumeMessahe: no PerUIDServer for %s", gregor.UIDFromMessage(m))
```
maxtaco: typo in string^

rpc/server/main.go
```diff
@@ -498,39 +430,56 @@ func (s *Server) publish(m gregor1.Message) error {
...
-		s.log.Error("publishCh full: %d", len(s.publishCh))
+		s.log.Error("publishDispatchCh full: %d", len(s.publishDispatchCh))
 		s.stats.Count("publish queue full")
 		return ErrPublishChannelFull
 	}
 	return nil
 }
 
-// Serve starts the serve loop for Server.
-func (s *Server) Serve() error {
```
maxtaco: what took the place of this?

rpc/server/user_server.go
mmaxim: Slice would suffice, I can change it.

rpc/server/user_server.go
mmaxim: Yep!

rpc/server/main.go
mmaxim: Yep

rpc/server/main.go
mmaxim: Sorry!

rpc/server/main.go
mmaxim: Nothing, the only thing this did was serialize access to `Server.users` which is handled by the lock now.

mmaxim:
@patrickxb I'm not sure the tests do a good job with that, the close tests are fairly easily to pass. I can toss together something that is harder (maybe I can get it to fail on master), but I think this patch does it 100% right.

rpc/server/main.go
```diff
@@ -253,12 +228,13 @@ func (s *Server) waitOnUIDServer(usrv *perUIDServer) {
 		select {
 		case <-usrv.ShouldShutdown():
 			s.Lock()
-			if usrv.Shutdown(false) {
+			if usrv.ConfirmShutdown() {
```
maxtaco: how about:
```
s.Lock()
confirmed := usrv.ConfirmShutdown()
if confirmed {
   delete(s.users, s.uidKey(usrv.uid))
}
s.Unlock()
if confirmed {
    usrv.Shutdown()
    if s.events != nil {
         // ...
    }
}
```
I get very nervous about unlocking a lock conditionally along two different branches, it's easy to mess that up in future additions...

mmaxim: Ok that works!

maxtaco:
I had one suggestion for cleanup, otherwise 👍

maxtaco:
🚢

PR #123, opened 2016-06-27
add GetItem() to State CORE-3137

mmaxim:
@maxtaco this is a PR for the gregor changes in the CORE-3137 on the client.


protocol/gregor1/factory.go
```diff
@@ -259,4 +263,22 @@ func (o ObjFactory) MakeReminderSetFromReminders(reminders []gregor.Reminder, mo
 	return ret, nil
 }
 
+func (o ObjFactory) ExportState(s gregor.State) (gregor.State, error) {
+
+	if gs, ok := s.(State); ok {
```
maxtaco: When would this not work/

interface.go
```diff
@@ -208,13 +209,15 @@ type ObjFactory interface {
...
 	MakeState(i []Item) (State, error)
+	MakeStateWithLookupTable(i []Item, table map[string]Item) (State, error)
 	MakeMetadata(uid UID, msgid MsgID, devid DeviceID, ctime time.Time, i InBandMsgType) (Metadata, error)
 	MakeInBandMessageFromItem(i Item) (InBandMessage, error)
 	MakeMessageFromInBandMessage(i InBandMessage) (Message, error)
 	MakeTimeOrOffsetFromTime(t time.Time) (TimeOrOffset, error)
 	MakeTimeOrOffsetFromOffset(d time.Duration) (TimeOrOffset, error)
 	MakeReminderSetFromReminders([]Reminder, bool) (ReminderSet, error)
 	UnmarshalState([]byte) (State, error)
+	ExportState(s State) (State, error)
```
maxtaco: this feels a bit like an abuse of the `ObjFactory` concept, since it's pushing objects in the opposite direction. Can you dangle this off of `type State interface` or does that not work?

mmaxim: I tried that initially and forget why I ended up here, let me try again now that things are a little different.

mmaxim:
Point of the last commit is to move `Export` onto `State`, and to have it return a different type than `State`. Got rid of other `ObjFactory`, and I think this makes things simpler.

protocol/gregor1/extras.go
```diff
@@ -271,6 +280,14 @@ func (s State) Hash() ([]byte, error) {
 	return sum[:], nil
 }
 
+func (s State) Export() (gregor.ProtocolState, error) {
+	return s, nil
+}
+
+func (s State) ProtocolName() string {
```
maxtaco: why is this needed?

mmaxim: I wanted to have `ProtocolState` be different than `State` to make `Export` more meaningful (and have it such that `FastLookupState` can't be returned from `Export`), so I just added this. This particular function isn't really used anywhere. `ProtocolState` is also not necessary, but I thought it was kind of a nice way to indicate what is going on in interfaces.go, where you can usually only reference `State`.

maxtaco:
👍

PR #125, opened 2016-06-28
use the DB's NOW() if possible

maxtaco:
- on consumption of dismissals, we were sometimes using a time that was inserted into the DB and read back out, which
  in some cases wound up rounding up.


maxtaco:
Closed in favor of a better fix.

PR #128, opened 2016-07-01
add new State RPC and change dismissal behavior in InBandMessages Since CORE-3289

mmaxim:
@maxtaco 

This makes two key changes:

1.) Add a new `State` RPC that is now used during sync, specifically in the case where we fail due to a hash mismatch, as well as if we are doing a fresh sync (from 0).
2.) Change how dismissals are processed. Now, we filter out dismissals that are applied to items that are not present on the receiver of `InBandMessagesSince`. This is more consistent with the optimization to also remove the creation items for this same case.

We can walk through this tomorrow when you are free.


storage/sql_sm.go
```diff
@@ -525,20 +526,33 @@ func (s *SQLEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) *time.Time {
 	return ctime.TimeOrNil()
 }
 
+func (s *SQLEngine) isDismissed(msg gregor.InBandMessage, c time.Time) bool {
+	if update := msg.ToStateUpdateMessage(); update != nil {
+		if creation := update.Creation(); creation != nil {
+			if dtime := creation.DTime(); dtime != nil {
+				if t := dtime.Time(); t != nil {
+					return t.Before(c)
```
maxtaco: isn't there a utility to check this for `TimeOrOffset`?

rpc/client/client.go
```diff
@@ -123,19 +123,76 @@ func (c *Client) syncFromTime(cli gregor1.IncomingInterface, t *time.Time) (msgs
...
+		return msgs, err
+	}
+
 	return msgs, nil
 }
 
+func (c *Client) Sync(cli gregor1.IncomingInterface) ([]gregor.InBandMessage, error) {
+	var zerotime time.Time
+
+	latestCtime := c.sm.LatestCTime(c.user, c.device)
```
maxtaco: is it possible to use `time.Time.IsZero` for this?

mmaxim: Hah, yes... Sorry didn't know about that.

rpc/client/client.go
```diff
@@ -123,19 +123,76 @@ func (c *Client) syncFromTime(cli gregor1.IncomingInterface, t *time.Time) (msgs
...
+	for _, i := range items {
+		if ibm, err := c.sm.ObjFactory().MakeInBandMessageFromItem(i); err == nil {
+			res = append(res, ibm)
+		}
+	}
+	return res, nil
+}
+
+func (c *Client) State(cli gregor1.IncomingInterface) (gregor.State, error) {
+	ctx, _ := context.WithTimeout(context.Background(), 5*time.Second)
```
maxtaco: is it possible to pass this in as a configuration variable?

rpc/server/main.go
```diff
@@ -519,12 +532,12 @@ func (s *Server) storageSync(sm gregor.StateMachine, log rpc.LogOutput, arg greg
 
 // storageStateByCategoryPrefix schedules a query of the state filtered by
 // category prefix on the dispatch handler thread.
-func (s *Server) storageStateByCategoryPrefix(arg gregor1.StateByCategoryPrefixArg) stateByCategoryPrefixRes {
-	respCh := make(chan stateByCategoryPrefixRes)
-	req := stateByCategoryPrefixReq{arg: arg, respCh: respCh}
+func (s *Server) storageState(req stateReq) stateRes {
```
maxtaco: change the comment to match

storage/dismissal_filter.go
```diff
@@ -0,0 +1,52 @@
...
+		// that the message was created before the time we are selecting on
+		include := false
+		for _, dMsgID := range dismissal.MsgIDsToDismiss() {
+			if dMsg, present := allmsgs[dMsgID.String()]; present {
+				if dMsg.Metadata().CTime().Before(ctime) {
+					include = true
+					break
+				}
+			} else {
+				// If we don't have the message, assume that it is create before
```
maxtaco: `create` -> `created` in comments

storage/sql_sm.go
mmaxim: Maybe? If you could help me find it I would use it.

rpc/client/client.go
mmaxim: Sync has a timeout that is a second long, so I just added this to be consistent with it, seemed like overkill to thread it through configuration.

maxtaco: Ok.  What happens if we hit the timeout?  Is there an eventual retry?

mmaxim: No, there is no no retry on the sync process if it doesn't work (on any kind of error).

maxtaco: Hm, that's a little scary, because in the case of a slight hiccup, there will be unrecoverable and bad behavior.  I think we should either: not timeout, or do a retry in the case of a timeout, or maybe bump the timeout way up

storage/sql_sm.go
```diff
@@ -525,20 +526,33 @@ func (s *SQLEngine) LatestCTime(u gregor.UID, d gregor.DeviceID) *time.Time {
...
-               i.category, i.body,
+               i.category, i.body, i.dtime,
                dt.category, dt.dtime,
                di.dmsgid
 	        FROM gregor_messages AS m
 	        LEFT JOIN gregor_items AS i ON (m.uid=i.UID AND m.msgid=i.msgid)
 	        LEFT JOIN gregor_dismissals_by_time AS dt ON (m.uid=dt.uid AND m.msgid=dt.msgid)
 	        LEFT JOIN gregor_dismissals_by_id AS di ON (m.uid=di.uid AND m.msgid=di.msgid)
-	        WHERE m.uid=? AND (i.dtime IS NULL OR i.dtime > `
+	        WHERE m.uid=? `
```
maxtaco: i worry this will wind up grabbing a whole bunch of messages from the DB just to throw them away.  why not filter both ways?

maxtaco: Also, relatedly, we should probably have a sweeper that goes through and deletes old dismissed messages.  Actually, do we already have this?

rpc/server/main.go
mmaxim: I'll bump both timeouts, but it's probably worth thinking about the retry story for syncing.

storage/sql_sm.go
mmaxim: I'll swing by to talk about this.

maxtaco: ok 👍

maxtaco:
OK 👍 LgTM!

test/helpers.go
```diff
@@ -438,3 +482,70 @@ func AddReminder(sm gregor.StateMachine, d time.Duration) gregor.Reminder {
...
+
+	var ref []gregor.InBandMessage
+	var msgs []gregor1.InBandMessage
+	logMsg := func(ibm gregor.InBandMessage) {
+		t.Logf("msgID: %s Ctime: %s", ibm.Metadata().MsgID(), ibm.Metadata().CTime())
+	}
+	msgs = append(msgs, NewRandomCreation(uid, basetime))
+
+	t.Logf("--- msgs for dismissal ---")
+	ref = append(ref, msgs[len(msgs)-1])
```
maxtaco: what is `ref`?  comment better for future readability

maxtaco: can you make a helper function that pushes onto `msgs` always and `ref` sometimes with a flag to better self-document what's going on

test/helpers.go
```diff
@@ -438,3 +482,70 @@ func AddReminder(sm gregor.StateMachine, d time.Duration) gregor.Reminder {
 	}
 	return rm
 }
+
+func checkMessages(t *testing.T, res []gregor.InBandMessage, ref []gregor.InBandMessage) {
```
maxtaco: some docs as to what this guy is for too

PR #131, opened 2016-07-04
fix several gregor recent gregor regressions in stateByCategoryPrefix

maxtaco:
- burned by the nil interface v nil type bug on deviceID and timeOrOffset, giving us nonsensical query parameters
- another regression -- since we are threading clockwork's real clock through more code, we were usually not using NOW() on mysql, which gave us inaccurate time stamps.  I'm not sure when that bug snuck in, but i came across it as i was debugging queries.


rpc/server/main.go
```diff
@@ -601,11 +601,18 @@ func (s *Server) storageDispatchHandler() {
 		case stateReq:
 			var res gregor.State
 			var err error
+			var deviceID gregor.DeviceID
+			if req.deviceID != nil && len(req.deviceID) > 0{
+				deviceID = req.deviceID
```
maxtaco: You hit the *worst* feature of Go.  Welcome to the club!

mmaxim:
LGTM!

maxtaco:
Btw here is the bug in play: https://golang.org/doc/faq#nil_error

maxtaco:
We really need a test for this.  I'll add it in a separate PR.

PR #132, opened 2016-07-05
tests for #131

maxtaco:


maxtaco:
BTW, confirmed that this test breaks before #131

mmaxim:
👍

PR #134, opened 2016-07-05
Record the build hash in the docker image

jzila:


bin/Dockerfile
```diff
@@ -17,5 +17,6 @@ WORKDIR /home/keybase
 COPY gregor.sh /home/keybase/gregor.sh
 COPY gregord/gregord /home/keybase/gregord
 COPY dbinit/dbinit /home/keybase/dbinit
+COPY revision .
```
maxtaco: I see where this file is written, but where is it read back out?

jzila: It isn't yet, we'd echo that in the kbweb build.

maxtaco:
ok 👍 !

PR #136, opened 2016-07-08
authenticate version rpc calls CORE-3352

mmaxim:
@maxtaco 

I'll have a PR on keybase/keybase soon for the authd side of this.


maxtaco:
👍

PR #137, opened 2016-07-14
only use data from Restore() if it is newer than a day CORE-3299

mmaxim:
@maxtaco r?

The purpose of this is to not try and sync down a down of IBMs from the server if the result of `Restore` is older than a day. It could be the state is blank and is a much smaller payload. Didn't seem necessary to make it configurable, since the user would almost certainly never set it.


rpc/client/client.go
```diff
@@ -73,7 +73,19 @@ func (c *Client) Restore() error {
 		return err
 	}
 
-	return c.sm.InitState(state)
+	if err := c.sm.InitState(state); err != nil {
+		return err
+	}
+
+	// If the latest CTime is more than a day old, then nuke this saved state and fetch from the server fresh
```
maxtaco: This makes a lot of assumption about the rate of message arrival.  If we're only getting ~5 messages a day, and there are ~200 days worth of activity, this seems like the wrong idea, doesn't it?  

I would choose a much longer timeout here.  I wonder if the correct idea is to set the timeout adaptively based on estimates of how many messages come in, and how many have come in previously.

maxtaco:
I had a deep question about this feature, but otherwise, the code looks coode

rpc/client/client.go
mmaxim: Hmm, now that you mention it I think this feature may not be worth it. The main goal here was to bring the state down in the cases where you might end up syncing a ton of old creations and dismissals. It turns out though that `InBandMessagesSince` already is smart about not sending those messages down anyway (this is partially a result of my change to not bring down dismissals of creations in the future). I don't think I fully understood that when I made this ticket, and didn't rethink it before implementing.

I can just this out, or make the timeout like a week, whatever you think works best.

maxtaco: Let's discuss, but probably worth backing out if you don't think there's a need for it.

PR #140, opened 2016-07-21
Slack send on error and cleanup after build

jzila:


jzila:
I'm going to merge since this has already been reviewed in other repositories.

PR #145, opened 2016-07-26
add option to connect to logstash linked container

mmaxim:
@maxtaco r?

This makes it possible to link to a container called "logstash" on ECS, and have gregor connect to it on a syslog interface. Can't say I really tested this since it is so tightly coupled to AWS, I'll have to try it there, but it is close enough to the Loggly config that I think we should be good.


bin/log.go
```diff
@@ -79,6 +79,25 @@ var formatMetaData = `%{time:15:04:05.000} ▶ %{level:.4s} %{program} %{shortfi
...
+func getSyslogAddr() (string, string) {
+	trySyslog := func(prefix string) string {
+		syslogIP := os.Getenv(prefix + "_ADDR")
+		syslogPort := os.Getenv(prefix + "_PORT")
+		if syslogIP != "" && syslogPort != "" {
+			return syslogIP + ":" + syslogPort
+		}
+		return ""
+	}
+	if addr := trySyslog("LOGSTASH_PORT_514_TCP"); addr != "" {
```
maxtaco: is it weird that `LOGSTASH_PORT_514_TCP_PORT=515` is  a thing?

mmaxim: I guess, it's just the difference between the internal/external ports of the container displayed in an awkward way in that variable name.

maxtaco:
ok, 👍

PR #146, opened 2016-07-28
retry syslog connection CORE-3429

mmaxim:
@maxtaco r?

The purpose of this patch is to try the syslog connection a set amount of times before giving up, as opposed to just exiting if the first try fails. This is essential for logstash to work on ECS, since it takes like 5 seconds for Java (logstash is written in Java) to start up and bind to it's port in the linked container.


maxtaco:
👍

